PR #482, opened 2023-12-23
Keep TypeReference<List<Step>?> object

meladRaouf:


feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/cache/OrchestratorCache.kt
```diff
@@ -54,3 +55,6 @@ internal class OrchestratorCache @Inject constructor(
         }
     }
 }
+
+@Keep
+internal object StepsTypeReference : TypeReference<List<Step>?>()
```
BurningAXE, line 60: Can we keep TypeReference directly so we don't need to do this for each generic type?
https://stackoverflow.com/questions/28433281/how-to-include-typereference-proguard-rule

feature/alert/src/main/java/com/simprints/feature/alert/AlertResult.kt
```diff
@@ -8,7 +8,4 @@ import java.io.Serializable
 data class AlertResult(
     val buttonKey: String,
     val payload: Bundle,
-) : Serializable {
-
-    fun isBackButtonPress() = buttonKey == AlertContract.ALERT_BUTTON_PRESSED_BACK
```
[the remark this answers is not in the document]
meladRaouf: @BurningAXE Android studio flagged it as unused code

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/cache/OrchestratorCache.kt
meladRaouf: @BurningAXE This is a good idea.
But I don't like the suggested solution in Stackoverflow the `-keepattributes Signature`  The solution will keep all signatures not only Type references

feature/alert/src/main/java/com/simprints/feature/alert/AlertResult.kt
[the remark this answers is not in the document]
BurningAXE: I'd still double check that.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/cache/OrchestratorCache.kt
BurningAXE: I don't see any issue with that - doubt it will increase code side by much. And we don't have anything to hide now :)

meladRaouf: I used this rule to keep all type reference subclasses 

```
-keep class * extends com.fasterxml.jackson.core.type.TypeReference {
    public <init>(java.lang.reflect.Type);
}
```

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/cache/OrchestratorCacheIntegrationTest.kt
```diff
@@ -21,7 +21,6 @@ import org.junit.Ignore
 import org.junit.Test
 import org.junit.runner.RunWith
 
-@Ignore("Jackson behaves differently in testing than in production, to be fixed later")
```
meladRaouf, line 24: There is an issue when serializing the results data 
I enabled this test so that we fix the issue before merging the PR

alexandr-simprints:
Comments on the [06f9317](https://github.com/Simprints/Android-Simprints-ID/pull/482/commits/06f9317779d93a8d28d0337d24465af0d8b633d8) changes above.

`OrchestratorCache`
When `OrchestratorCache::steps` are serialized into JSON using `jsonHelper.toJson(value)`, it created a following structure:
```
["Arrays$ArrayList":[{ /*step1json*/ }, { /*step2json*/ }]]
```
Due to this `"Arrays$ArrayList"` specification, it cannot be unmarshalled using the `object : TypeReference<List<Step>>() {}` type.

*Fix:*
Since it's only necessary to create a JSON array with the Step objects, we can do it explicitly in a following way:
```
val jsonArray = List<Step>.joinToString(separator = ",") {
    jsonHelper.toJson(it, module = stepsModule)
  }.let { "[$it]" }

// jsonArray = [{ /*step1json*/ }, { /*step2json*/ }, etc.]
```
Then the  `object : TypeReference<List<Step>>() {}` type can be used to unmarshal the steps correctly.

alexandr-simprints (approved):

BurningAXE:
> Comments on the [06f9317](https://github.com/Simprints/Android-Simprints-ID/pull/482/commits/06f9317779d93a8d28d0337d24465af0d8b633d8) changes above.
> 
> `OrchestratorCache` When `OrchestratorCache::steps` are serialized into JSON using `jsonHelper.toJson(value)`, it created a following structure:
> 
> ```
> ["Arrays$ArrayList":[{ /*step1json*/ }, { /*step2json*/ }]]
> ```
> 
> Due to this `"Arrays$ArrayList"` specification, it cannot be unmarshalled using the `object : TypeReference<List<Step>>() {}` type.
> 
> _Fix:_ Since it's only necessary to create a JSON array with the Step objects, we can do it explicitly in a following way:
> 
> ```
> val jsonArray = List<Step>.joinToString(separator = ",") {
>     jsonHelper.toJson(it, module = stepsModule)
>   }.let { "[$it]" }
> 
> // jsonArray = [{ /*step1json*/ }, { /*step2json*/ }, etc.]
> ```
> 
> Then the `object : TypeReference<List<Step>>() {}` type can be used to unmarshal the steps correctly.

Nice hack! I fought with this but focused on trying to fix the `toJson` conversion, instead of going around it.

PR #488, opened 2023-12-27
CORE-3160 Check camera permission

meladRaouf:


alexandr-simprints (approved):
While this works as a quick workaround, I think that we need to design the application more thoughtfully, and consider all the possible states of the camera permissions:
> Allowed

Initialize the camera and preview.

> May ask the camera permission

Display the permission dialog. 
- If allowed, initialize the camera and preview.
- If rejected, display the rationale message and an action button that opens the permission dialog once again.
- If rejected and 'Don't ask again' is selected, display the rationale message asking user to enable the camera permissions in settings, since the dialog can no longer be displayed. Add an action button that navigates to the app's settings page.

> Cannot ask for permissions, because 'Don't ask again' was previously selected

Display the rationale message asking user to enable the camera permissions in settings, since the dialog can no longer be displayed. Add an action button that navigates to the app's settings page.

PR #493, opened 2024-01-03
[CORE-3167] Schedule or cancel firmware worker along with other periodic workers

BurningAXE:
See ticket for description of the problem. Fix is to simply schedule (or cancel) the firmware worker along with the other periodic workers - currently happening at app startup.


id/build.gradle.kts
```diff
@@ -36,11 +36,14 @@ dependencies {
     implementation(project(":feature:orchestrator"))
     implementation(project(":feature:dashboard"))
 
+    implementation(project(":fingerprint:infra:scanner"))
```
meladRaouf, line 39: @BurningAXE , are there any other options instead of adding a scanner dependancy in the ID module?

meladRaouf (approved):

id/build.gradle.kts
BurningAXE: Yeah, didn't like it either. However, the firmware worker is (for obvious reasons) inside the scanner module. If we want to avoid this dependency we can move the scheduling to another place (maybe the Orchestrator) and do it on every workflow. However, I don't like this solution for two reasons:
1. It will scatter worker scheduling in different places, making it harder to reason about workers, etc
2. Similar to why we moved the worker scheduling to app startup - if the user doesn't trigger any workflows for a while it will lead to the firmware worker incorrectly (not) triggering.

So, if there isn't any hard requirement not to make `id` depend on the scanner module, I think this is the lesser of the two evils. Feel free to suggest anything not evil, though :D

id/src/main/java/com/simprints/id/ScheduleBackgroundSyncUseCase.kt
```diff
@@ -21,6 +27,9 @@ class ScheduleBackgroundSyncUseCase @Inject constructor(
             imageUpSyncScheduler.scheduleImageUpSync()
             configManager.scheduleSyncConfiguration()
             authManager.scheduleSecurityStateCheck()
+            externalScope.launch {
```
luhmirin-s, line 30: IIRC, the convention for the coroutines is to make things simple on the callers' side and ensure correct schedulers in the lowest layer possible (inside of the scheduler in this case).

It would also simplify the testing.

BurningAXE: Yep, that would be better. Will update.

luhmirin-s (approved):

PR #494, opened 2024-01-04
Rename signing_properties.gradle so that it can be add the git ignore

meladRaouf:
I renamed the signing_properties.gradle.kts so that Git ignores the new name and stops annoying us.
after merging this PR you should completely delete `signing_properties.gradle.kts  `and create a new file named `signing_info.gradle.kts` 
I will also open the same PR in the main branch to simplify your life.

### DIFF
--- a/ci/pipeline/setup
+++ b/ci/pipeline/setup
@@ -4,3 +4,11 @@ echo "org.gradle.caching=true" >>gradle.properties
 echo "org.gradle.jvmargs=-Xmx7g"
 touch local.properties
 touch build-logic/signing_info.gradle.kts
+# Add dummy signing info to build-logic/signing_info.gradle.kts
+cat << EOF >> build-logic/signing_info.gradle.kts
+extra.apply {
+    set("store_file", "dummy")
+    set("store_password", "dummy")
+    set("key_alias", "dummy")
+    set("key_password", "dummy")
+}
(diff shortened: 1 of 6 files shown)


luhmirin-s (commented):
Shouldn't this PR also delete the file from the repo?

luhmirin-s:
Also, IMO this branch should be in main only. Nothing to do with the release process or hotfixes.

meladRaouf:
> Shouldn't this PR also delete the file from the repo?

@luhmirin-s I deleted it now

meladRaouf:
> Also, IMO this branch should be in main only. Nothing to do with the release process or hotfixes.

I wanted to open this PR in both branches so that we can forget the signing config hassle.

ci/pipeline/setup
```diff
@@ -3,3 +3,12 @@ chmod +x gradlew
 echo "org.gradle.caching=true" >>gradle.properties
 echo "org.gradle.jvmargs=-Xmx7g"
 touch local.properties
+touch build-logic/signing_info.gradle.kts
+# Add dummy signing info to build-logic/signing_info.gradle.kts
+cat << EOF >> build-logic/signing_info.gradle.kts
+extra.apply {
+    set("store_file", "dummy")
```
luhmirin-s, line 10: Wouldn't this break debug key signing?

luhmirin-s (approved):

ci/pipeline/setup
meladRaouf: No it won't break anything. I added the dummey data for the unit test and sonarqube steps to pass.
The real signing data is being added by the `ci/deployment/signing_properties_setup` script

alexandr-simprints (approved):

BurningAXE (approved):

PR #496, opened 2024-01-04
[CORE-3169] Comply with Firebase limitations for event names and messages length

BurningAXE:
This started out with a benign notice of a log from production SID that Configuration worker's tag is incompatible with Firebase Analytics' limitations (see related ticket). Then I kept going into rabbit holes finding more and more instances where our logs didn't make it through to Firebase because they were either too long or contained invalid characters. As a result here are a few different changes:

- ConfigurationWorker's tag was renamed
- Some tags were also renamed (replacing spaces with underscores)
- Some worker logs were stripped of params since they were too long to fit
- Simber was updated to ensure that even if its users don't comply with the rules. In release/staging it makes the tags/messages valid by replacing or trimming the strings. In debug this throws an exception so we can notice and fix the problem rather than keep using a modified tag/message

Open question: how do we make sure this problem doesn't repeat in the future?
We didn't notice it so far because in debug we don't "plant" Firebase Crashlytics/Analytics so we don't get any logs in logcat if we don't comply with their rules. I can think of a few possible solutions:

1. Have a step during release testing where we intentionally connect a release version to AS and monitor logcat. This is relatively simple but doesn't ensure we'll run over all possible logs that can happen.
2. Create and use debug projects of Firebase Crashlytics/Analytics so we have a more realistic debug app but still don't pollute the prod dashboards
3. Anything better?

### DIFF
--- a/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
+++ b/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
@@ -119,10 +119,35 @@ object Simber {
      * 1 kB in size."
      */
     fun tag(tag: String, isUserProperty: Boolean = false): Simber {
-        if (isUserProperty)
-            Timber.tag(USER_PROPERTY_TAG + tag)
-        else
-            Timber.tag(tag)
+        var conformingTag = tag
+
+        var maxLength = 24
+        if (isUserProperty) {
+            conformingTag = USER_PROPERTY_TAG + conformingTag
+            maxLength = 40
+        }
+
+        // Ensure length is not greater than allowed by Firebase Analytics
+        if (conformingTag.length > maxLength) {
+            if (BuildConfig.DEBUG) {
+                throw IllegalArgumentException("Tag must be less than 40 characters.")
+            } else {
+            conformingTag = conformingTag.substring(0, maxLength)
+            }
+        }
+
+        // Check that tag complies with Firebase Analytics requirements:
+        // Name must consist of letters, digits or _ (underscores).
+        if (tag.contains(Regex("[^a-zA-Z0-9_]"))) {
+            // Throw an exception in debug but replace invalid characters in other modes
+            if (BuildConfig.DEBUG) {
+                throw IllegalArgumentException("Tag must consist of letters, digits or _ (underscores).")
+            } else {
+                conformingTag = tag.replace(Regex("[^a-zA-Z0-9_]"), "_")
+            }
+        }
+
+        Timber.tag(conformingTag)
         return Simber
     }
 
--- a/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
+++ b/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
@@ -16,6 +16,9 @@ object Simber {
 
     internal const val USER_PROPERTY_TAG = "zzUserPropertyTag"
     private const val FIREBASE_ANALYTICS_MAX_MESSAGE_LENGTH = 100
+    private const val FIREBASE_ANALYTICS_MAX_TAG_LENGTH = 24
+    private const val FIREBASE_ANALYTICS_MAX_USER_TAG_LENGTH = 40
+    private val firebaseInvalidCharactersRegex = Regex("[^a-zA-Z0-9_]")
 
     /**
      * Use this when you want to go absolutely nuts with your logging. If for some reason you've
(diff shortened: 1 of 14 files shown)


infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -119,13 +138,51 @@ object Simber {
...
         return Simber
     }
 
+    /*
+    * Check that tag complies with Firebase Analytics requirements:
+    * Name must consist of letters, digits or _ (underscores).
+     */
+    private fun ensureCharactersAreValid(tag: String): String {
+        var validTag = tag
+        if (validTag.contains(Regex("[^a-zA-Z0-9_]"))) {
```
luhmirin-s, line 163: You are compiling 2 regex instances per call. I would suggest extracting the regex out of the method entirely and reusing the same instance.

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -119,13 +138,51 @@ object Simber {
...
+            if (BuildConfig.DEBUG) {
+                throw IllegalArgumentException("Tag must consist of letters, digits or _ (underscores).")
+            } else {
+                validTag = validTag.replace(Regex("[^a-zA-Z0-9_]"), "_")
+            }
+        }
+        return validTag
+    }
+
+    private fun ensureMaxLength(message: String, maxLength: Int): String {
```
luhmirin-s, line 174: My first instinct when seeing this name is that it pads message always to be precisely maxLength. :D Maybe something like `limitLength(message, max)`?

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -119,13 +138,51 @@ object Simber {
...
+        if (message.length > maxLength) {
+            if (BuildConfig.DEBUG) {
+                throw IllegalArgumentException("String must be less than $maxLength characters.")
+            }
+
+            return message.substring(0, maxLength)
```
luhmirin-s, line 180: I recommend using `String.take(maxLength)` as it does not require the length check and, therefore, the whole method could be simplified. 
Also, I would suggest trimming even in debug mode to make it a bit more noticeable while developing/debugging.

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -119,13 +138,51 @@ object Simber {
...
      */
     fun tag(tag: String, isUserProperty: Boolean = false): Simber {
-        if (isUserProperty)
-            Timber.tag(USER_PROPERTY_TAG + tag)
-        else
-            Timber.tag(tag)
+        var conformingTag = tag
+
+        // Firebase Analytics requires tag to be no longer than 24 characters
+        var maxLength = 24
```
luhmirin-s, line 144: This could be simplified a bit.

```
var conformingTag = ensureCharactersAreValid(tag)
conformingTag = if (isUserProperty) {
   ensureMaxLength(conformingTag, 24) // TODO extract const
} else {
   ensureMaxLength(USER_PROPERTY_TAG + conformingTag, 40) // TODO extract const
}
Timber.tag(conformingTag)
```

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
luhmirin-s: Also, the variable could be avoided since you can call `contains()` on the tag param and return the result of `replace()` or the initial tag as-is directly.

luhmirin-s (approved):

meladRaouf (approved):

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
BurningAXE: Currently in debug mode it crashes so it's unavoidably noticeable :D

alexandr-simprints (approved):

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -16,6 +16,9 @@ object Simber {
 
     internal const val USER_PROPERTY_TAG = "zzUserPropertyTag"
     private const val FIREBASE_ANALYTICS_MAX_MESSAGE_LENGTH = 100
+    private const val FIREBASE_ANALYTICS_MAX_TAG_LENGTH = 24
+    private const val FIREBASE_ANALYTICS_MAX_USER_TAG_LENGTH = 24
```
luhmirin-s, line 20: Shouldn't this be 40?

BurningAXE: Ooops!

PR #503, opened 2024-01-11
[CORE-2207] Show foreground notification while executing workers (Syncing)

alex-vt:
Workers in the app are set to foreground, with a "Sync in progress" low-priority notification while work in progress.

System restrictions apply - see https://developer.android.com/develop/background-work/services/foreground-services#bg-access-restrictions. The workers in the app do not appear to be eligible to exemptions from restrictions, so the notification visibility to user will be limited: on new Android versions by default when the app UI was on screen when the worker started.

On new Android versions:
* User sees a permission system prompt to enable notifications, in the same flow as the location permission system prompt. Permissions may often be declined.
* Restrictions for notifications appearing while the app is in the background could be removed by setting battery saving option in App Info system menu to Unrestricted. Again, user would have to be promted to do that, wth a limited rate of success. The nature of these sync workers may not be important enough to bother user with that.
* Restrictions could be lifted for work done in exact alarm broadcast receivers. The sync workers by their nature appear the to far less time critical than an alarm clock app would be though; the app may have difficulties at Play Store reviews with such implementation - probably not a way to go.

String resource needing translation: `infra/resources/src/main/res/values/strings.xml` `notification_sync_description` "Sync in progress".


infra/core/src/main/java/com/simprints/core/workers/SimCoroutineWorker.kt
```diff
@@ -61,4 +115,11 @@ abstract class SimCoroutineWorker(
             }
         }
     }
+
+    private companion object {
+        private const val WORKER_FOREGROUND_NOTIFICATION_ID = 2
+        private const val WORKER_FOREGROUND_NOTIFICATION_CHANNEL_ID =
+            "WORKER_FOREGROUND_NOTIFICATION"
+        private const val WORKER_FOREGROUND_NOTIFICATION_CHANNEL_NAME = "Sync in progress"
```
luhmirin-s, line 123: Shouldn't this also be a localised string since it is shown in a few places in the system UI?

luhmirin-s (approved):

PR #504, opened 2024-01-11
CORE-3184 Enable gzip for post request bodies

luhmirin-s:


alex-vt (approved):

infra/network/src/main/java/com/simprints/infra/network/httpclient/DefaultOkHttpClientBuilder.kt
```diff
@@ -65,11 +72,55 @@ internal class DefaultOkHttpClientBuilder {
...
+
+    private class CompressedRequestBody(private val requestBody: RequestBody) : RequestBody() {
+
+        override fun contentType(): MediaType? = requestBody.contentType()
+
+        override fun contentLength(): Long = -1 // We don't know the compressed length in advance!
+
+        @Throws(IOException::class)
+        override fun writeTo(sink: BufferedSink) {
+            GzipSink(sink).buffer().use { gzipSink -> requestBody.writeTo(gzipSink) }
```
BurningAXE, line 123: Is the default compression level the one that was agreed upon in the sync revamp doc?

BurningAXE (approved):

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/remote/EnrolmentRecordApiInterface.kt
```diff
@@ -3,11 +3,13 @@ package com.simprints.infra.enrolment.records.store.remote
 import com.simprints.infra.enrolment.records.store.remote.models.ApiEnrolmentRecords
 import com.simprints.infra.network.SimRemoteInterface
 import retrofit2.http.Body
+import retrofit2.http.Headers
 import retrofit2.http.POST
 import retrofit2.http.Path
 
 internal interface EnrolmentRecordApiInterface : SimRemoteInterface {
 
+    @Headers("Content-Encoding: gzip")
```
ethiery, line 12: FYI we have not added support for content encoding on this endpoint on our side.
But it makes sense, I'll add it early next week.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/EventRemoteInterface.kt
```diff
@@ -37,6 +38,7 @@ internal interface EventRemoteInterface : SimRemoteInterface {
         @Query("lastEventId") lastEventId: String?
     ): ResponseBody
 
+    @Headers("Content-Encoding: gzip")
```
ethiery, line 41: FYI we have not added support for content encoding on this endpoint on our side.
But it makes sense, I'll add it early next week.

infra/network/src/main/java/com/simprints/infra/network/httpclient/DefaultOkHttpClientBuilder.kt
ethiery, line 123: Do you have any idea what compression level this is using by default?

ethiery (approved):

infra/network/src/main/java/com/simprints/infra/network/httpclient/DefaultOkHttpClientBuilder.kt
luhmirin-s: It is either none or some auto-detecting one

Screenshot 2024-01-15 at 09 19 42

luhmirin-s: Explicitly set the compression level to 6.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/remote/EnrolmentRecordApiInterface.kt
luhmirin-s: @ethiery wouldn't it be simple to just blank enable gzip support across endpoints based on the headers?

infra/network/src/main/java/com/simprints/infra/network/httpclient/DefaultOkHttpClientBuilder.kt
luhmirin-s: Manually set to level 6.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/remote/EnrolmentRecordApiInterface.kt
ethiery: It would not make much difference complexity wise on BFSID's side. 
I think the only reason I did not propose a blanket compression is that compressing small payloads is most likely counter-productive (as in, the latency added by SID compressing and BFSID decompressing will exceed the latency decrease caused by the tiny bandwidth savings). But it's hard to say for sure without actually benchmarking it.

ethiery: If it's all the same to you, I'd rather not fix anything that is not broken and keep the compression/decompression for endpoints that see large request/response bodies.

luhmirin-s: I didn't mean it as compress everything. Just having the capability to accept compressed payloads if the `Content-Encoding` header is present. You would not need to think about it for any future changes.

On SID side that is a capability built into the OkHttp by default (for response bodies).

PR #511, opened 2024-01-17
CORE-3187 Project sync worker revamp

luhmirin-s:
* Moved cancellation of scheduled work out of auth-logic to allow doing it from config-sync as well (when the project ended or the device compromised)
* Use the project state field from config instead of the old device endpoint (device endpoint revamp coming in next PR).

Note that parts in the auth-logic are commented out to avoid compilation errors and failing tests but preserved for reference when implementing CORE-3188. It will be deleted in the upcoming PR.


infra/config-sync/src/main/java/com/simprints/infra/config/sync/usecase/HandleProjectStateUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+            authManager.signOut()
+        }
+    }
+
+    private suspend fun shouldSignOut(projectId: String, projectState: ProjectState): Boolean {
+        val isProjectEnded = projectState == ProjectState.PROJECT_ENDED
+        val isProjectEnding = projectState == ProjectState.PROJECT_ENDING
+        val hasNoEventsToUpload = eventSyncManager.countEventsToUpload(projectId, null).first() == 0
+
+        return isProjectEnded || (isProjectEnding && hasNoEventsToUpload)
```
ybourgery, line 37: If the project is ENDED, you should try to sync the record and then log out even if the sync fails

ybourgery (changes requested):

infra/config-sync/src/main/java/com/simprints/infra/config/sync/usecase/HandleProjectStateUseCase.kt
luhmirin-s: @ybourgery Is that a new requirement? This check mirrors the one in the old "secure state" worker also it seems that BFSID would rejected records for ended projects - https://simprints.atlassian.net/browse/CORE-2277

ybourgery: Ah yes sorry you are right it's the Confluence doc that is out-of-date

ybourgery (approved):

infra/config-sync/src/main/java/com/simprints/infra/config/sync/usecase/HandleProjectStateUseCase.kt
```diff
@@ -0,0 +1,40 @@
...
+internal class HandleProjectStateUseCase @Inject constructor(
+    private val configScheduler: ProjectConfigurationScheduler,
+    private val securityStateScheduler: SecurityStateScheduler,
+    private val imageUpSyncScheduler: ImageUpSyncScheduler,
+    private val eventSyncManager: EventSyncManager,
+    private val authManager: AuthManager
+) {
+
+    suspend operator fun invoke(projectId: String, state: ProjectState) {
+        if (shouldSignOut(projectId, state)) {
```
ybourgery, line 21: Are you not dealing with the `PAUSED` state of the project, where you can only up-sync?

luhmirin-s: There is a check in the event sync worker that disables down sync (it was previously checking the secure state).

PR #512, opened 2024-01-17
CORE-3188 Device state sync endpoint

luhmirin-s:
* Implemented device config sync worker to check compromised state and schedule record upsync (as per design doc).
* Deleted everything related to "secure state" as it has been replaced with explicit project state and "isCompromised" flag in device endpoint.


infra/config-sync/src/main/java/com/simprints/infra/config/sync/usecase/HandleProjectStateUseCase.kt
```diff
@@ -1,31 +1,18 @@
...
-    private val configScheduler: ProjectConfigurationScheduler,
-    private val securityStateScheduler: SecurityStateScheduler,
-    private val imageUpSyncScheduler: ImageUpSyncScheduler,
     private val eventSyncManager: EventSyncManager,
-    private val authManager: AuthManager
+    private val logoutUseCase: LogoutUseCase,
 ) {
 
     suspend operator fun invoke(projectId: String, state: ProjectState) {
         if (shouldSignOut(projectId, state)) {
```
ybourgery, line 14: If the project is `ENDED`, you should try to sync the record and then log out even if the sync fails

ybourgery (changes requested):

ybourgery (approved):

alexandr-simprints (approved):

PR #516, opened 2024-01-18
[CORE-1022] Replace deprecated Apis startActivityForResult() and onActivityResult()

alex-vt:
The only remaining usage of deprecated onActivityResult() and onActivityResult() was for `GooglePlayServicesAvailabilityChecker` in `GoogleApiAvailability.showErrorDialogFragment`.

Updated to a newer variant of `GoogleApiAvailability.showErrorDialogFragment`, that uses `ActivityResultLauncher` instead.

### DIFF
--- a/feature/login/src/main/java/com/simprints/feature/login/screens/form/LoginFormFragment.kt
+++ b/feature/login/src/main/java/com/simprints/feature/login/screens/form/LoginFormFragment.kt
@@ -1,10 +1,12 @@
 package com.simprints.feature.login.screens.form
 
-import android.content.Intent
 import android.os.Bundle
 import android.view.View
 import android.widget.Toast
 import androidx.activity.addCallback
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.IntentSenderRequest
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.annotation.StringRes
 import androidx.core.view.isVisible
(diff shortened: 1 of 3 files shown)


feature/login/src/main/java/com/simprints/feature/login/screens/form/LoginFormFragment.kt
```diff
@@ -50,6 +51,17 @@ internal class LoginFormFragment : Fragment(R.layout.fragment_login_form) {
     private val binding by viewBinding(FragmentLoginFormBinding::bind)
     private val viewModel by viewModels<LoginFormViewModel>()
 
+    private lateinit var checkForPlayServicesResultLauncher: ActivityResultLauncher<IntentSenderRequest>
+    init {
```
luhmirin-s, line 55: You can skip the init block and just declare it inline. It would be identical in execution, but a bit less disjointed.

luhmirin-s (approved):

feature/login/src/main/java/com/simprints/feature/login/screens/form/LoginFormFragment.kt
alex-vt: There's a reference to `checkForPlayServicesResultLauncher` inside `registerForActivityResult(ActivityResultContracts.StartIntentSenderForResult()) { ... }` block, so an inlined declaration wouldn't work.

PR #519, opened 2024-01-22
CORE-3192 Sync count simplification in workers

luhmirin-s:
Main changes: 
* Use a header instead of a parallel request to fetch the number of events to downsync.
* Dropping redundant worker for event up sync counting (up sync happens in single worker and can do just a single DB read at the beginning).
* Cleanup of now unnecessary workers.
* If the sync event count exceeds the BE threshold, switch the sync progress indicator to undetermined.


infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/EventRemoteDataSource.kt
```diff
@@ -51,15 +52,25 @@ internal class EventRemoteDataSource @Inject constructor(
...
+                ?.toBoolean() ?: false // If not present, assume it's not lower bound
+
+            val streaming = response.body()?.byteStream() ?: ByteArrayInputStream(byteArrayOf())
             Simber.tag("SYNC").d("[EVENT_REMOTE_SOURCE] Stream taken")
 
-            scope.produce(capacity = CHANNEL_CAPACITY_FOR_PROPAGATION) {
-                parseStreamAndEmitEvents(streaming, this)
-            }
+            EventDownSyncResult(
+                totalCount = totalCount.takeUnless { isTotalLowerBound },
```
BurningAXE, line 69: Why not take `totalCount` if it was sent but is lower bound?

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/workers/EventDownSyncDownloaderWorker.kt
```diff
@@ -117,3 +136,12 @@ internal suspend fun WorkInfo.extractDownSyncProgress(eventSyncCache: EventSyncC
     val cached = eventSyncCache.readProgress(id.toString())
     return maxOf(progress, output, cached)
 }
+
+internal suspend fun WorkInfo.extractDownSyncMaxCount(eventSyncCache: EventSyncCache): Int {
+    val progress = this.progress.getInt(PROGRESS_DOWN_MAX_SYNC, -1)
+    val output = this.outputData.getInt(OUTPUT_DOWN_MAX_SYNC, -1)
```
BurningAXE, line 142: Consts can probably be named a little more clearly - PROGRESS_DOWN_SYNC & MAX_DOWN_SYNC or something similar.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/EventUpSyncHelperImpl.kt
BurningAXE, line 1: This should be deleted?

BurningAXE (approved):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/EventUpSyncHelperImpl.kt
luhmirin-s: This was an empty file.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/EventRemoteDataSource.kt
luhmirin-s: It messes up the UI slightly. The problem is that if there are more items to down sync header returns (there are technical reasons for it on BE), we cannot meaningfully show the percent indicator and show a continuous one instead. That is the same not having the count in the first place, so why bother passing it around :D

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/workers/EventDownSyncDownloaderWorker.kt
luhmirin-s: I would say that it is a matter of preference. I wanted to keep the prefix as similar as possible to the non-max constants. It would also imply that similar constant should be used together all the time.
Since those constants are only relevant in the context of this file, I will leave it as is for now.

PR #531, opened 2024-01-25
Fix CommCare IDENTIFY response

BurningAXE:
Currently pointing to release/2023.4.0 but will change to release/2023.4.1 once the branch is ready.

- [ ] Point to `release/2023.4.1`

### DIFF
--- a/feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
+++ b/feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
@@ -35,7 +35,6 @@ internal class AlertFragment : Fragment(R.layout.fragment_alert) {
         super.onViewCreated(view, savedInstanceState)
 
         val config = args.alertConfiguration
-        Simber.tag(ALERT.name).i("Payload:  ${config.color.name} - ${getPayloadContent(config.payload)}")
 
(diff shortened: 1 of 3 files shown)


feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
alex-vt, line 1: Is the payload not needed in logs anymore?

alex-vt (approved):

feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
BurningAXE: It may be beneficial but unfortunately Firebase has a limit of 100 characters and this is too big to fit. So I replaced it with logging just the title to indicate an alert was shown. We can then match the same session in BFSID and see all the details.

meladRaouf:
> Currently pointing to release/2023.4.0 but will change to release/2023.4.1 once the branch is ready.
> 
>     * [ ]  Point to `release/2023.4.1`

As we haven't yet fully deployed 23.4.0, it is ok to merge this fix to 23.4.0, and no need to create 24.4.1

meladRaouf (approved):

luhmirin-s (approved):

PR #537, opened 2024-01-31
[MS-19] use root detection while syncing data

meladRaouf:
If the user could root his device after login, he could down sync biometric records from the backend.
I added a root detection check before doing any sync operations.

### DIFF
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
@@ -7,10 +7,15 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.simprints.core.domain.tokenization.TokenizableString
 import com.simprints.feature.dashboard.settings.syncinfo.modulecount.ModuleCount
+import com.simprints.infra.authstore.AuthStore
+import com.simprints.infra.config.store.ConfigRepository
 import com.simprints.infra.config.store.models.DownSynchronizationConfiguration
 import com.simprints.infra.config.store.models.ProjectConfiguration
 import com.simprints.infra.config.store.models.SynchronizationConfiguration
+import com.simprints.infra.config.store.models.TokenKeyType
 import com.simprints.infra.config.store.models.isEventDownSyncAllowed
+import com.simprints.infra.config.store.tokenization.TokenizationProcessor
+import com.simprints.infra.enrolment.records.store.EnrolmentRecordRepository
 import com.simprints.infra.enrolment.records.store.domain.models.SubjectQuery
 import com.simprints.infra.events.event.domain.models.EventType
 import com.simprints.infra.eventsync.EventSyncManager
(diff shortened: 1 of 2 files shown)


feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -146,6 +149,12 @@ internal class SyncInfoViewModel @Inject constructor(
     }
 
     private fun load() = viewModelScope.launch {
+        // if device is not safe stop all sync operations
+        // users can root the device after login and we need to stop sync operations
+        if (isDeviceSafe().not()) {
+            return@launch
```
ybourgery, line 155: Should we not add a specific message on the dashboard when this is the case, as it would be easier to debug if it happens in a project?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -146,6 +149,12 @@ internal class SyncInfoViewModel @Inject constructor(
     }
 
     private fun load() = viewModelScope.launch {
+        // if device is not safe stop all sync operations
+        // users can root the device after login and we need to stop sync operations
+        if (isDeviceSafe().not()) {
```
ybourgery, line 154: The background sync will still work no?

ybourgery (changes requested):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
meladRaouf: @ybourgery, which dashboard? 
if you mean the SID dashboard, I don't think we should notify the user that the sync operations are disabled becasue we is rooted the device.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
meladRaouf: @ybourgery  No, I chose the easiest way to disable the sync.
The load method calculates the needed info to be up or down-synced. if we break from this method nothing will be synced

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
ybourgery: Yes, the SID dashboard. If we don't notify it and a user is complaining because their device is not syncing, I'm not sure we could find that the root cause is because the device is rooted

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
meladRaouf: @ybourgery  I found I should also stop sync in EventSyncMasterWorker. will implement it now

meladRaouf:
Will close this one

PR #539, opened 2024-01-31
[MS-19] use root detection while syncing data

meladRaouf:
If the user roots his device after login, he could downsync biometric records from the backend.
I added a root detection check before doing any sync operations. also displayed a big alert screen in one dashboard screen to let the user knows that SID doesn't work in rooted devices.

### DIFF
--- a/feature/dashboard/src/main/res/layout/fragment_main.xml
+++ b/feature/dashboard/src/main/res/layout/fragment_main.xml
@@ -60,4 +60,44 @@
 
         </LinearLayout>
     </ScrollView>
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:id="@+id/rootedDeviceError"
+        android:visibility="gone"
+        android:background="@color/simprints_red">
+        <TextView
+            android:id="@+id/alertTitle"
+            style="@style/Text.Headline5.White.Bold"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:layout_marginBottom="24dp"
+            android:gravity="center"
+            android:maxLines="2"
+            app:layout_constraintBottom_toTopOf="@id/alertImage"
+            app:layout_constraintEnd_toEndOf="parent"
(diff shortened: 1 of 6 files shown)


ybourgery (approved):

feature/dashboard/src/main/res/layout/fragment_main.xml
```diff
@@ -60,4 +60,44 @@
...
         </LinearLayout>
     </ScrollView>
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:id="@+id/rootedDeviceError"
+        android:visibility="gone"
+        android:background="@color/simprints_red">
+        <TextView
+            android:id="@+id/alertTitle"
```
BurningAXE, line 70: If the TextView and ImageView will have IDs better to have clearer ones.

BurningAXE (approved):
Should we add such a check to the image sync worker, too?

feature/dashboard/src/main/res/layout/fragment_main.xml
meladRaouf: @BurningAXE Ok I will remove the IDs, I copied them from the alert screen :D

luhmirin-s: Can you use the alert screen for this? Similar to how it is handled in the regular flows (in the login-check module)

meladRaouf: @luhmirin-s We only show the alert screen in the workflows, not in the dashboard
I tried to use it and found it complex to import as the alert screen stores alert events, which is not supported outside the intent workflows.

meladRaouf:
> Should we add such a check to the image sync worker, too?

I don't see a need to stop any upsync worker. 
If the device is rooted we should stop it from down-syncing any biodata

BurningAXE:
> > Should we add such a check to the image sync worker, too?
> > @BurningAXE
> > I don't see a need to stop any upsync worker.
> > If the device is rooted we should stop it from down-syncing any biodata

With these changes we are preventing both down and upsync for event data, though.

PR #543, opened 2024-02-01
fix  getUserId issue // To be merged after 2023.3.4 go prod 100%

meladRaouf:
I added the fix here so that we don't miss it.


alex-vt (changes requested):
The unit test probably should be ported too (replace `GROUP` with `Partitioning` there).

alex-vt (approved):

infra/event-sync/src/test/java/com/simprints/infra/eventsync/status/down/EventDownSyncScopeRepositoryTest.kt
alex-vt, line 1: For reference: https://github.com/Simprints/Android-Simprints-ID/pull/538/files#diff-5ef858eb5a58e8a3c5f41eba2263985fd80e986141fb2252b1767341a65fd6e1

alex-vt (changes requested):

alex-vt (approved):

BurningAXE (approved):

alexandr-simprints (approved):

luhmirin-s:
I think this is redundant since we added a fix for MS-201 into the next release.

meladRaouf:
This fix was already added to  2023.4.1

PR #547, opened 2024-02-05
MS-201 Save user ID on login/flow to correctly sync user-scoped data

luhmirin-s:
Cherry-picked contents of #545 onto the release branch.

### DIFF
--- a/feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/UpdateStoredUserIdUseCase.kt
+++ b/feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/UpdateStoredUserIdUseCase.kt
@@ -1,0 +1,18 @@
+package com.simprints.feature.logincheck.usecases
+
+import com.simprints.core.domain.tokenization.TokenizableString
+import com.simprints.infra.authstore.AuthStore
+import javax.inject.Inject
+
+class UpdateStoredUserIdUseCase @Inject constructor(
+    private val authStore: AuthStore,
+) {
+
+    operator fun invoke(userId: TokenizableString) {
+        // This is a hack to make sure that there is a stored user ID if user logged in before 2024.1.0.
+        // It could be removed after all we drop support for older versions.
+        if (authStore.signedInUserId == null) {
+            authStore.signedInUserId = userId
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/UpdateStoredUserIdUseCase.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.simprints.infra.authstore.AuthStore
+import javax.inject.Inject
+
+class UpdateStoredUserIdUseCase @Inject constructor(
+    private val authStore: AuthStore,
+) {
+
+    operator fun invoke(userId: TokenizableString) {
+        // This is a hack to make sure that there is a stored user ID if user logged in before 2024.1.0.
+        // It could be removed after all we drop support for older versions.
```
alexandr-simprints, line 13: Do you think there might be a need to check the SID version as well? If no, feel free to resolve

infra/auth-store/src/main/java/com/simprints/infra/authstore/domain/LoginInfoStore.kt
```diff
@@ -23,6 +29,26 @@ internal class LoginInfoStore @Inject constructor(
...
+                prefs.getBoolean(USER_ID_TOKENIZED, false) -> TokenizableString.Tokenized(value)
+                else -> TokenizableString.Raw(value)
+            }
+        }
+        set(value) = prefs.edit {
+            if (value == null) {
+                remove(USER_ID_VALUE)
+                remove(USER_ID_TOKENIZED)
+            } else {
+                putBoolean(USER_ID_TOKENIZED, value is TokenizableString.Tokenized)
```
alexandr-simprints, line 46: `value is TokenizableString.Tokenized` can be changed to `value.isTokenized()`

alexandr-simprints (approved):

feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/UpdateStoredUserIdUseCase.kt
luhmirin-s: It isn't needed explicitly. As long as there is a cached user ID we do not care about the exact version.

BurningAXE (approved):

PR #548, opened 2024-02-06
MA-11 Use secure shared prefs in login info store

luhmirin-s:
* Switch LoginInfoStore to use encrypted shared prefs with data migration on the first data access.
* Some minor related cleanup - deleted redundant method in AuthStore/LoginInfoStore and deleted useless TODOs.

Note that this change adds a bit of startup time - on my Sony F8331 with Android 8 cold app start from launcher to dashboard changes from around 6-7s to 8-9s.

### DIFF
--- a/infra/auth-store/src/main/java/com/simprints/infra/authstore/domain/LoginInfoStore.kt
+++ b/infra/auth-store/src/main/java/com/simprints/infra/authstore/domain/LoginInfoStore.kt
@@ -129,8 +129,4 @@ internal class LoginInfoStore @Inject constructor(
             remove(CORE_FIREBASE_API_KEY)
         }
     }
-
-    fun storeCredentials(projectId: String) {
-        signedInProjectId = projectId
-    }
 }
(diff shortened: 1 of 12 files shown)


meladRaouf (approved):

infra/auth-store/src/main/java/com/simprints/infra/authstore/domain/LoginInfoStore.kt
```diff
@@ -26,86 +30,105 @@ internal class LoginInfoStore @Inject constructor(
...
+    private val prefs: SharedPreferences =
+        ctx.getSharedPreferences(LEGACY_PREF_FILE_NAME, Context.MODE_PRIVATE)
+
+    private val securePrefs = securityManager.buildEncryptedSharedPreferences(SECURE_PREF_FILE_NAME)
+
+    /**
+     * Ensures that data has been migrated to secure prefs before accessing it.
+     */
+    private fun getSecurePrefs(): SharedPreferences {
+        if (prefs.contains(PROJECT_ID)) {
```
BurningAXE, line 43: Wonder if this would have a non-negligible impact on performance - each time a value is read or saved, the OS will open two files instead of one. Maybe we can reverse this and check if `PROJECT_ID` is present in secure prefs and only proceed to migration (and hence opening the legacy file) if it's missing.

luhmirin-s: I was wondering the same and implemented "in-memory" cache for project and user ID fields, assuming that those probably are the hottest paths. There was no perceivable difference in startup times and app work in general. 

I think the main overhead comes from creating/opening the encrypted file, then the reference is stored in memory (since this is a singleton) and the read/writes are pretty quick.

PR #549, opened 2024-02-06
MS-161-coSync-1-N

BurningAXE:


luhmirin-s:
@BurningAXE I suggest closing this for now and reopening after rebase and update

BurningAXE:
> @BurningAXE I suggest closing this for now and reopening after rebase and update

@luhmirin-s I just rebased it yesterday and started working on it :D

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepository.kt
```diff
@@ -1,9 +1,32 @@
...
 
+import com.simprints.core.ExcludedFromGeneratedTestCoverageReports
 import com.simprints.infra.config.store.models.Project
+import com.simprints.infra.enrolment.records.store.domain.models.BiometricDataSource
+import com.simprints.infra.enrolment.records.store.domain.models.FaceIdentity
+import com.simprints.infra.enrolment.records.store.domain.models.FingerprintIdentity
+import com.simprints.infra.enrolment.records.store.domain.models.SubjectQuery
 import com.simprints.infra.enrolment.records.store.local.EnrolmentRecordLocalDataSource
 
+@ExcludedFromGeneratedTestCoverageReports("This is an interface with no logic")
```
luhmirin-s, line 11: Technically, the default values could be considered business logic.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
```diff
@@ -109,4 +115,30 @@ internal class EnrolmentRecordRepositoryImpl(
             project = project
         )
     }
+
+    override suspend fun count(query: SubjectQuery, dataSource: BiometricDataSource): Int =
```
luhmirin-s, line 119: Does it still make sense to have the delegation in the declaration? Should the repo even implement the local data source interface in the first place?

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
```diff
@@ -109,4 +115,30 @@ internal class EnrolmentRecordRepositoryImpl(
...
+        when (dataSource) {
```
luhmirin-s, line 120: I would suggest something like:
```
fun fromIdentitySource(dataSource: BiometricDataSource): IdentityDataSource = 
    when (dataSource) {
       BiometricDataSource.SIMPRINTS -> localDataSource
       BiometricDataSource.COMMCARE -> commCareDataSource
    }
```

And then on callers side could be simplified to 
```
enrolmentRecordRepo.fromIdentitySource(dataSource).<method>()
```

infra/enrolment-records-store/src/test/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSourceTest.kt
```diff
@@ -0,0 +1,404 @@
...
...projectId\":\"nXcj9neYhXP9rFp56uWk\",\"moduleId\":{\"value\":\"AWuA3H0WGtHI2uod+ePZ3yiWTt9etQ==\"},\"attendantId\":{\"value\":\"AdySMrjuy7uq0Dcxov3rUFIw66uXTFrKd0BnzSr9MYXl5maWEpyKQT8AUdcPuVHUWpOkO88=\"},\"biometricReferences\":[{\"id\":\"2b9b4991-29d7-3eee-ac02-191afaa0c1a2\",\"templates\":[{\"quality\":99,\"template\":\"Rk1SACAyMAAAAADkAAABLAGQAMUAxQEAABBjIYCzAAgQAEAhABoeAIB3ACQNAEAoACsYAEDvADYDAECrAFgDAEBQAHIaAEDUAH2AAEEEAIX4AEDhAJX4AEDpALZ7AIB1AL0RAEEJAMPuAECSAM0NAED2ANPxAEDmAOXsAEDdAPrnAEBhAQYaAIDcARDaAICwARn8AEDJASlnAECMATcVAEB7ATgbAECxAT2qAECnAT6hAEB3AT6eAECIAT6eAEEWAUDGAIDHAUizAED9AU2zAEDsAVU9AED1AWYzAIC2AWciAAAA\",\"finger\":\"LEFT_THUMB\"},{\"quality\":88,\"template\":\"Rk1SACAyMAAAAAEUAAABLAGQAMUAxQEAABBYKUBqACiHAIBKAD0RAECyAEjuAEAfAF6sAEAnAG4sAIB4AHT2AIByAIuiAEAGAIs1AEC9AJDUAEB3AJS9AICaAJXRAEDzAJ/bAECHAKi9AIBzALJDAEAXALfDAEB6ALszAIDGAL3JAED2AL/QAEBHAMrRAEENAMrWAECiAMy9AIAkANfNAEDwANnKAEB3AOr4AICjAPKhAEANAPXGAIEEAPjJAEBCAPzaAICNAQUDAEEZAQdpAEDOARsfAIDpAR0TAEDZASUUAEDnASeNAIDsAS0NAED+ATONAEDxAT8KAIB9AUB5AEDAAVUNAEBdAVZzAECCAWt9AAAA\",\"finger\":\"LEFT_INDEX_FINGER\"}],\"format\":\"ISO_19794_2\",\"type\":\"FINGERPRINT_REFERENCE\"}]},\"type\":\"EnrolmentRecordCreation\"}]}"
```
luhmirin-s, line 42: With long blobs of JSON it is much better to use the triple quote string as it does not require you to escape every quotemark and you can also just copy-paste the string contents.

infra/enrolment-records-store/src/test/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSourceTest.kt
```diff
@@ -0,0 +1,404 @@
...
+        val actualIdentities = dataSource.loadFingerprintIdentities(query, range)
+
+        assertEquals(expectedFingerprintIdentities.size, actualIdentities.size)
+        val areContentsEqual =
+            expectedFingerprintIdentities
+                .zip(actualIdentities) { expected, actual ->
+                    expected.subjectId == actual.subjectId &&
+                            expected.fingerprints
+                                .zip(actual.fingerprints) { expectedFingerprint, actualFingerprint ->
+                                    expectedFingerprint.fingerIdentifier ==
```
luhmirin-s, line 173: Breaking the comparison like that really hurts readability :(

infra/enrolment-records-store/src/test/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSourceTest.kt
```diff
@@ -0,0 +1,404 @@
...
+                listOf("someOtherDatumId", "subjectActions", "someOtherDatumId", "subjectActions")
+        every { mockDataCursor.getString(1) } returnsMany
+                listOf(SUBJECT_ACTIONS_FACE_1, SUBJECT_ACTIONS_FACE_2)
+
+        val query = SubjectQuery()
+        val range = 0..expectedFaceIdentities.size
+        val actualIdentities = dataSource.loadFaceIdentities(query, range)
+
+        assertEquals(expectedFaceIdentities.size, actualIdentities.size)
+        val areContentsEqual =
```
luhmirin-s, line 207: Is this auto-formatting? It is really hard to read due to those new lines and spaces.

feature/client-api/src/test/java/com/simprints/feature/clientapi/mappers/request/requestFactories/EnrolActionFactory.kt
```diff
@@ -23,6 +23,7 @@ internal object EnrolActionFactory : RequestActionFactory() {
         userId = MOCK_USER_ID.asTokenizableRaw(),
         metadata = MOCK_METADATA,
         moduleId = MOCK_MODULE_ID.asTokenizableRaw(),
+        biometricDataSource = "",
```
luhmirin-s, line 26: Could this be set to some mock value to align with other params?

luhmirin-s (commented):
Added some things to consider. 

I find the new version of `EnrolmentRecordRepositoryImpl` a bit confusing - it delegates most of the methods in a non-transparent way, but I don't have any concrete suggestions for now, more of a prompt for a discussion.

BurningAXE:
> Added some things to consider.
> 
> I find the new version of `EnrolmentRecordRepositoryImpl` a bit confusing - it delegates most of the methods in a non-transparent way, but I don't have any concrete suggestions for now, more of a prompt for a discussion.

Why do you think it's delegating in a non-transparent way? (We may take this to Slack if you think it will be long discussion)

fingerprint/infra/simprints-bio-sdk/src/main/java/com/simprints/fingerprint/infra/biosdkimpl/matching/CrossMatching.kt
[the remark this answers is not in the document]
BurningAXE: Weird! I definitely did not add it intentionally but was probably an artifact of some other branch I switched from.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepository.kt
BurningAXE: Do you think it merits a test?

id/src/main/AndroidManifest.xml
```diff
@@ -9,6 +9,8 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
 
+    <uses-permission android:name="org.commcare.dalvik.provider.cases.read" />
```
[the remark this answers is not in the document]
BurningAXE: Since access is done in `infra/enrolment-records-store` would it make sense to put the permission it its manifest? This way if we refactor modules there wouldn't be a hidden dependency on other modules.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
BurningAXE: On the first question - do you mean to move the current implementation from the repo to the interface?
On the second - it doesn't need it actually. Will remove.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
BurningAXE: It would simplify the repo implementation a little but would expose the IdentityDataSource directly to callers. Currently we simply forward the calls but having it pass through the repo reserves the right to alter the logic in the future. So not sure it's worth it.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -0,0 +1,163 @@
...
+                        template = encoder.base64ToBytes(faceTemplate.template),
+                        format = faceReference.format,
+                    )
+                }
+            }
+        )
+    }
+
+    private fun loadEnrolmentRecordCreationEvents(range: IntRange): List<EnrolmentRecordCreationEvent> {
+        val enrolmentRecordCreationEvents: MutableList<EnrolmentRecordCreationEvent> = mutableListOf()
```
alexandr-simprints, line 80: Would be better to use [Kotlin's use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) for cursor operations to ensure the proper closure of the resources once done or exception is caught. Currently, the cursor isn't closed when the exception is thrown.

```
try {
    context.contentResolver.query(CASE_METADATA_URI, null, null, null, null).use {
        if (caseMetadataCursor != null && caseMetadataCursor.moveToPosition(range.first)) { ... }
    }
} catch { ... }
```

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    private fun loadEnrolmentRecordCreationEvents(caseId: String): List<EnrolmentRecordCreationEvent> {
+        val caseEnrolmentRecordCreationEvents: MutableList<EnrolmentRecordCreationEvent> = mutableListOf()
+
+        //Access Case Data Listing for the caseId
+        val caseDataUri = CASE_DATA_URI.buildUpon().appendPath(caseId).build()
+        val caseDataCursor =
+            context.contentResolver.query(caseDataUri, null, null, null, null)
+        var subjectActions = ""
```
alexandr-simprints, line 112: I'd also utilized the [Kotlin's use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/) inline function here

infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEnrolmentRecordEvents.kt
```diff
@@ -1,9 +1,9 @@
-package com.simprints.feature.clientapi.models
+package com.simprints.infra.events.event.cosync
 
 import androidx.annotation.Keep
 import com.simprints.infra.events.event.domain.models.subject.EnrolmentRecordEvent
 
 @Keep
-internal data class CoSyncEnrolmentRecordEvents(
+data class CoSyncEnrolmentRecordEvents(
```
alexandr-simprints, line 7: Is the removal of the `internal` keyword intentional?

infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEvents.kt
```diff
@@ -1,9 +1,9 @@
-package com.simprints.feature.clientapi.models
+package com.simprints.infra.events.event.cosync
 
 import androidx.annotation.Keep
 import com.simprints.infra.events.event.domain.models.Event
 
 @Keep
-internal data class CoSyncEvents(
+data class CoSyncEvents(
```
alexandr-simprints, line 7: Is the removal of the `internal` keyword intentional?

alexandr-simprints (changes requested):

infra/enrolment-records-store/src/test/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSourceTest.kt
BurningAXE: Yeah, sorry. Will reformat.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/domain/models/BiometricDataSource.kt
```diff
@@ -0,0 +1,13 @@
+package com.simprints.infra.enrolment.records.store.domain.models
+
+enum class BiometricDataSource {
+    SIMPRINTS,
+    COMMCARE;
+
+    companion object {
+        fun fromString(value: String) = when (value) {
+            "COMMCARE" -> COMMCARE
```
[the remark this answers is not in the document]
BurningAXE: It's up to us really to define the API but I will add a `.uppercase()` to avoid any confusion.

infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEvents.kt
BurningAXE: Same for this one.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepository.kt
luhmirin-s: I don't think so, just pointing out a technicality 🤓

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
luhmirin-s: I know it does, but should it?

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
luhmirin-s: IMO, such DSL-like API could be better, but that is a matter of taste. Although, you can still use private a method for matching internally to reduce duplication a bit.

luhmirin-s:
> Why do you think it's delegating in a non-transparent way? (We may take this to Slack if you think it will be long discussion)

If I see a method called on repo and then open the source I either don't see the method when opening the repo, or see a completely different class when opening the methods implementation. 

Now that I think about it, delegating this way also forces the local store interface to be public (which IMO is unnecessary).

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
luhmirin-s: Also, `IdentityDataSource` is currently public, so it is already exposed :)

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -0,0 +1,155 @@
...
+import com.simprints.infra.events.event.cosync.CoSyncEnrolmentRecordEvents
+import com.simprints.infra.events.event.domain.models.subject.EnrolmentRecordCreationEvent
+import com.simprints.infra.events.event.domain.models.subject.FaceReference
+import com.simprints.infra.events.event.domain.models.subject.FingerprintReference
+import com.simprints.infra.logging.Simber
+import com.simprints.libsimprints.Constants.SIMPRINTS_COSYNC_SUBJECT_ACTIONS
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+class CommCareIdentityDataSource @Inject constructor(
```
luhmirin-s, line 27: Should be at least internal.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/EnrolmentRecordRepositoryImpl.kt
BurningAXE: Not sure I get you - the local data source offers more functions as we also modify it unlike the CommCare one which we just read. So, in that regard we do need the extra functionality from the local data source.

BurningAXE:
> > Why do you think it's delegating in a non-transparent way? (We may take this to Slack if you think it will be long discussion)
> 
> If I see a method called on repo and then open the source I either don't see the method when opening the repo, or see a completely different class when opening the methods implementation.
> 
> Now that I think about it, delegating this way also forces the local store interface to be public (which IMO is unnecessary).

Ah, yes, I found this part a little confusing, too. Just to be clear - it was like this before my changes. Whether we should refactor it is an open question, though.

alexandr-simprints (approved):

PR #557, opened 2024-02-12
complete the setup if there are no required licenses.

meladRaouf:


.github/workflows/jira.yml
```diff
@@ -2,6 +2,8 @@
 name: Jira Sync
 
 on:
+    workflow_dispatch:
+
```
meladRaouf, line 6: We should have at least one trigger here.
this should be deleted once the jira integration is ready

luhmirin-s (approved):

BurningAXE (approved):

PR #558, opened 2024-02-12
MS-42 Session Event repository

luhmirin-s:
This is a preparation refactoring to clear the event's data structure and DB APIs before introducing a new set of events. Main changes:

* Introduces event scope types. All events and scopes currently fall under the "SESSION" type, but that will change soon.
* `sessionId` in events is replaced with a more explicit `scopeId`. This is because there will be events outside of a session. "Session ID" is now a subset of the more generic "Scope ID".
* Since the session scope type will remain the most important, business logic related to storing/updating "current session" has been moved to a separate `SessionEventRepository`. 
* At the same time, more generic operations in `EventRepository` have been pruned a bit. 
* Also there are some minor fixes in DB setup - fixed validator package declarations, removed unused DAO methods, removed interfaces for classes that are always internal, etc.

This may look like a lot of changes, but it is mostly due to moving some files, new database schema and tests.


infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScopeEndCause.kt
```diff
@@ -0,0 +1,16 @@
...
+
+@Keep
+internal enum class ApiEventScopeEndCause {
+
+    WORKFLOW_ENDED,
+    NEW_SESSION,
+}
+
+internal fun EventScopeEndCause?.fromDomainToApi(): ApiEventScopeEndCause = when (this) {
+    null, EventScopeEndCause.WORKFLOW_ENDED -> ApiEventScopeEndCause.WORKFLOW_ENDED
```
alexandr-simprints, line 14: From the business logic, is it correct to associate `null` with a finished workflow?

infra/events/src/main/java/com/simprints/infra/events/event/local/EventDatabaseFactory.kt
```diff
@@ -1,8 +1,55 @@
...
+            Simber.e(t)
+            throw t
+        }
+    }
+
+    private fun getOrCreateKey(@Suppress("SameParameterValue") dbName: String): CharArray {
+        return try {
+            securityManager.getLocalDbKeyOrThrow(dbName)
+        } catch (t: Throwable) {
+            t.message?.let { Simber.d(it) }
```
alexandr-simprints, line 37: Are we sure that we want to use debug-level logging in this scenario?

infra/events/src/main/java/com/simprints/infra/events/event/local/EventLocalDataSource.kt
```diff
@@ -1,25 +1,151 @@
...
+                    block()
+                } else {
+                    throw ex
+                }
+            }
+        }
+
+    private fun isFileCorruption(ex: Throwable) =
+        ex is SQLiteDatabaseCorruptException ||
+            ex.let { it as? SQLiteException }?.message?.contains("file is not a database") == true
```
alexandr-simprints, line 80: We might want to add `ignoreCase=true` here just in case

infra/events/src/main/java/com/simprints/infra/events/event/local/migrations/EventMigration13to14.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    companion object {
+
+        private const val EVENT_SCOPE_TABLE_NAME = "DbEventScope"
+        private const val SESSION_SCOPE_TABLE_NAME = "DbSessionScope"
+
+        private const val EVENT_SCOPE_COLUMNS =
+            "id, projectId, type, start_unixMs, start_isTrustworthy, start_msSinceBoot, end_unixMs, end_isTrustworthy, end_msSinceBoot, payloadJson"
+        private const val SESSION_SCOPE_OLD_COLUMNS =
+            "id, projectId, 'SESSION', start_unixMs, start_isTrustworthy, start_msSinceBoot, end_unixMs, end_isTrustworthy, end_msSinceBoot, payloadJson"
```
alexandr-simprints, line 49: I believe you tested this migration, but just want to make sure that the `'SESSION'` column name is written correctly here

alexandr-simprints (approved):
Approving. Please address a couple of minor notes. Feel free to resolve if not relevant

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScopeEndCause.kt
luhmirin-s: IMO, it is ok in this case since this is the default fallback value. The only other end cause is NEW_SESSION which is the one we actually care about.

infra/events/src/main/java/com/simprints/infra/events/event/local/EventDatabaseFactory.kt
luhmirin-s: I didn't really change that part. This file shows up in diff because I removed the redundant internal interface and this code was previously in "Impl" file.

infra/events/src/main/java/com/simprints/infra/events/event/local/EventLocalDataSource.kt
luhmirin-s: Same as the database factory -  this file shows up in diff because I removed the redundant internal interface and this code was previously in "Impl" file.

infra/events/src/main/java/com/simprints/infra/events/event/local/migrations/EventMigration13to14.kt
luhmirin-s: I checked the database contents after migration, and they were correct. To be clear, in this case, "SESSION" is the default value set to the "type" column for each event scope.

meladRaouf (approved):

PR #567, opened 2024-02-20
[MS-216] Consent text missing spaces - fix

alex-vt:
Sentences were concatenated without spaces separating them.

### DIFF
--- a/feature/consent/src/main/java/com/simprints/feature/consent/screens/consent/helpers/ParentalConsentTextHelper.kt
+++ b/feature/consent/src/main/java/com/simprints/feature/consent/screens/consent/helpers/ParentalConsentTextHelper.kt
@@ -42,12 +42,12 @@ internal class ParentalConsentTextHelper @Inject constructor(
         programName: String,
         modalityUseCase: String,
     ) = when (config?.enrolmentVariant) {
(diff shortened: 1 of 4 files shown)


feature/consent/src/main/java/com/simprints/feature/consent/screens/consent/helpers/ParentalConsentTextHelper.kt
```diff
@@ -117,4 +117,12 @@ internal class ParentalConsentTextHelper @Inject constructor(
         Modality.FINGERPRINT -> context.getString(R.string.consent_biometrics_access_fingerprint)
         else -> ""
     }
+
+    private fun StringBuilder.appendSentence(text: String): StringBuilder {
```
alexandr-simprints, line 121: Let's avoid the code duplication and move it to a single place of declaration

alexandr-simprints (changes requested):
`appendSentence` should be declared once to avoid code duplication

feature/consent/src/main/java/com/simprints/feature/consent/screens/consent/helpers/ParentalConsentTextHelper.kt
alex-vt: Sure - see https://github.com/Simprints/Android-Simprints-ID/pull/568

PR #570, opened 2024-02-21
2023.4.1+firebase fixes

meladRaouf:


meladRaouf:
> @meladRaouf There are two preceding commits in the issue that I don't see here: Screenshot 2024-02-21 at 20 17 33

Done

BurningAXE (approved):

PR #571, opened 2024-02-21
MS-214 Orchestrator to support configuration changes

alexandr-simprints:
These are optimal minimal changes required for the Orchestrator to support the orientation changes.
- Navigation stays in place.
- The application doesn't crash.
- Orchestrator's state seems to be consistent between the configuration changes.
-All other known minor issues are related to the children's fragments, not the orchestrator itself.


alex-vt (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
```diff
@@ -11,13 +12,21 @@ import com.simprints.infra.uibase.navigation.handleResult
 import com.simprints.infra.uibase.viewbinding.viewBinding
 import dagger.hilt.android.AndroidEntryPoint
 
+private const val KEY_IS_GRAPH_INITIALIZED = "KEY_IS_GRAPH_INITIALIZED"
```
luhmirin-s, line 15: I would prefer to keep constants in companion objects to keep class-per-file rules.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
```diff
@@ -28,13 +37,20 @@ internal class OrchestratorActivity : BaseActivity() {
...
     override fun onStart() {
         super.onStart()
-        val action = intent.action.orEmpty()
-        val extras = intent.extras ?: bundleOf()
-
-        findNavController(R.id.orchestrationHost).setGraph(
-            R.navigation.graph_orchestration,
-            OrchestratorFragmentArgs(action, extras).toBundle()
-        )
+        if(!isGraphInitialized) {
```
luhmirin-s, line 40: You could also do this to avoid tracking flags across configuration changes: 
```
with(findNavController(R.id.orchestrationHost)) {
            try {
                graph
            } catch (e: IllegalStateException) {
                val action = intent.action.orEmpty()
                val extras = intent.extras ?: bundleOf()

                setGraph(
                    R.navigation.graph_orchestration,
                    OrchestratorFragmentArgs(action, extras).toBundle()
                )
            }
        }
```

It is probably not better in terms of readability, though :D

luhmirin-s (approved):

meladRaouf (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
alexandr-simprints: IMO, a single flag state reads out way better than a try/catch block

PR #573, opened 2024-02-22
MS-205 Image upload halt when event sync starts

luhmirin-s:
Main changes:
* Added a listener for event sync master worker state changes and cancel+reschedule image upload worker.

Things I learned in the process:
* `isStopped` flag in `CoroutineWorker` is completely useless. Cancelling the worker cancels the whole coroutine scope, and the code within the scope does not get a chance to check if it should stop 🤷🏻 
* Having foreground notification in a worker prevents if from being cancelled by the work manager. It also ignores the `isStopped` flag :D 
* Cancelling a periodic worker by ID seems to prevent future workers from being rescheduled, contrary to what the internet says.

### DIFF
--- a/id/src/main/java/com/simprints/id/Application.kt
+++ b/id/src/main/java/com/simprints/id/Application.kt
@@ -16,6 +16,7 @@ import dagger.hilt.android.HiltAndroidApp
 import io.reactivex.exceptions.UndeliverableException
 import io.reactivex.plugins.RxJavaPlugins
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import javax.inject.Inject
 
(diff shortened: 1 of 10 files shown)


alex-vt (approved):

id/src/main/java/com/simprints/id/Application.kt
```diff
@@ -44,6 +45,11 @@ open class Application : CoreApplication(), Configuration.Provider {
         initApplication()
     }
 
+    override fun onTerminate() {
```
alex-vt, line 48: Is this mainly for debug purposes?

alex-vt (approved):

id/src/main/java/com/simprints/id/Application.kt
luhmirin-s: Added this to be 100% sure that the global scope is properly cancelled when the app is terminated.

meladRaouf (approved):

luhmirin-s:
Uncovered lines are just a convenience wrapper around worker manager API, and they are not excluded automatically due to sonar needing to understand how inline works.

PR #574, opened 2024-02-22
MS-229 Event upload batch size config

luhmirin-s:
Main changes:
* Added batch size configuration to `synchronization.up.simprints`. Defaulting all batch sizes to 1 if no config provided to mimic existing behaviour. 
* Using new configuration to upload events in the batches of provided sizes
* Split out-of-session events into separate request per scope to allow batching those events separately.

### DIFF
--- a/infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
+++ b/infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
@@ -181,11 +181,11 @@ internal class ConfigLocalDataSourceImpl @Inject constructor(
                         simprints = UpSynchronizationConfiguration.SimprintsUpSynchronizationConfiguration(
                             kind = UpSynchronizationConfiguration.UpSynchronizationKind.NONE,
                             batchSizes = UpSynchronizationConfiguration.UpSyncBatchSizes.default(),
+                            imagesRequireUnmeteredConnection = false,
                         ),
                         coSync = UpSynchronizationConfiguration.CoSyncUpSynchronizationConfiguration(
                             kind = UpSynchronizationConfiguration.UpSynchronizationKind.NONE
                         ),
-                        imagesRequireUnmeteredConnection = false,
                     ),
                     down = DownSynchronizationConfiguration(
                         partitionType = DownSynchronizationConfiguration.PartitionType.USER,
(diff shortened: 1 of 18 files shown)


ybourgery:
@luhmirin-s Do we agree on adding that we will add the new configuration parameter inside `synchronization.up.simprints` as it feels strange to have some in `synchronization.up` (imagesRequireUnmeteredConnection) and some other in `synchronization.up.simprints`?

luhmirin-s:
@ybourgery The discussion stopped with some agreement and no other comments, so I assumed we were doing it. I can hold merging PR until we are sure.
The location of the config makes sense to me. Maybe it is not too late to move the image imagesRequireUnmeteredConnection into `synchronization.up.simprints` to keep things consistent. Although I am not against moving this one layer up to `synchronization.up`.

luhmirin-s:
@ethiery Dragging you into this as well, do you think that we should move connection flag into `simprints` or move `batches` out of it?

infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
```diff
@@ -179,7 +179,12 @@ internal class ConfigLocalDataSourceImpl @Inject constructor(
                     frequency = SynchronizationConfiguration.Frequency.PERIODICALLY,
                     up = UpSynchronizationConfiguration(
                         simprints = UpSynchronizationConfiguration.SimprintsUpSynchronizationConfiguration(
-                            kind = UpSynchronizationConfiguration.UpSynchronizationKind.NONE
+                            kind = UpSynchronizationConfiguration.UpSynchronizationKind.NONE,
+                            batchSizes = UpSynchronizationConfiguration.UpSyncBatchSizes(
+                                sessions = 1,
```
alexandr-simprints, line 184: Why isn't `UpSyncBatchSizes.default()` being used here and in the `OldProjectConfig` below?

alexandr-simprints (approved):

ethiery:
> @ethiery Dragging you into this as well, do you think that we should move connection flag into `simprints` or move `batches` out of it?

Batching does not apply to both types of "up-sync". For co-sync, only sessions are "up-synced", one by one, through callbacks. Batching only applies to up-sync to simprints' back-end, so I would definitely nest `batches` under `simprints`.

And I'd say the same reasoning applies to image sync. Wifi/no wifi is irrelevant to co-sync since it does not happen over the internet, but locally through intents. So it probably would have made more sense for `ImagesRequireUnmeteredConnection` to be nested under `simprints`.

luhmirin-s:
@ybourgery @ethiery, it has not been released to prod, so we can change it pretty easily.
I will move the flag into Simprints' specific configuration on the SID side. Please do the same on BE side.

infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
luhmirin-s: 🤷🏻

PR #580, opened 2024-02-26
MS-215 Sync info screen revamp

luhmirin-s:
Main changes:
* Combine the "Records to download" and "Records to delete" fields in sync info since that info will be combined in the new events API. 
* Also made minor changes to the sync info layout to make it adapt to various configurations a bit better.
* Switched event fetch from dedicated endpoint to HEAD request on the events endpoint. Additionally added handling for "isLowerBound" handling on the sync info page.


feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -195,15 +185,15 @@ internal class SyncInfoViewModel @Inject constructor(
...
+            DownSyncCounts(0, false)
         }
 
     private suspend fun fetchAndUpdateRecordsToDownSyncAndDeleteCount(): DownSyncCounts =
         try {
             eventSyncManager.countEventsToDownload()
         } catch (t: Throwable) {
             Simber.d(t)
-            DownSyncCounts(0, 0)
+            DownSyncCounts(0, false)
```
alexandr-simprints, line 196: In the future, I'd like us to use the explicit parameter names, especially in cases like this. `DownSyncCounts(0, false)` doesn't make much sense when reading. 

In comparison, `DownSyncCounts(count = 0, isLowerBound = false)` is easier to understand. Or at least specifying the flag's purpose: `DownSyncCounts(0, isLowerBound = false)`

infra/resources/src/main/res/values/strings.xml
```diff
@@ -365,7 +365,7 @@
     <string name="dashboard_sync_info_title">Sync Information</string>
     <string name="dashboard_sync_info_select_modules_button">Select modules</string>
     <string name="dashboard_sync_info_records_to_upload">Records to upload</string>
-    <string name="dashboard_sync_info_records_to_download">Records to download</string>
+    <string name="dashboard_sync_info_records_to_download">Records to download or delete</string>
```
alexandr-simprints, line 368: Is this noted to be reflected in the other translations?

alexandr-simprints (approved):

PR #582, opened 2024-02-27
[MS-241] Add -keep declarations for all classes saved in the OrchestratorCache

BurningAXE:
FSR the @Keep annotation doesn’t work and in release builds the result classes are obfuscated which prevents Jackson from properly (de)serializing them and without the step results Enrol Last Bio cannot work.
To reproduce - either use a prod version or a debug one with isMinifyEnabled = true . Then log into a project with “Duplicate biometric check enabled” set to true and try to do an Identification followed by Enrol Last Bio. In 2023.4.2 it fails with a gray screen saying that there was a duplicate (another issue we need to fix later).

### DIFF
--- a/id/proguard-rules.pro
+++ b/id/proguard-rules.pro
@@ -52,6 +52,21 @@
 -keep class com.simprints.infra.events.event.domain.models.** { *; }
 -keep class com.simprints.infra.events.event.local.** { *; }
 
+-keep class com.simprints.feature.orchestrator.steps.** { *; }
+-keep class com.simprints.face.capture.** { *; }
+-keep class com.simprints.face.configuration.** { *; }
+-keep class com.simprints.feature.alert.** { *; }
+-keep class com.simprints.feature.consent.** { *; }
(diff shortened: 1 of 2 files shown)


meladRaouf (approved):

id/proguard-rules.pro
```diff
@@ -52,6 +52,21 @@
 -keep class com.simprints.infra.events.event.domain.models.** { *; }
 -keep class com.simprints.infra.events.event.local.** { *; }
 
+-keep class com.simprints.feature.orchestrator.steps.** { *; }
```
luhmirin-s, line 55: Wouldn't this completely disable obfuscation for those feature modules? Maybe we could have something more specific - e.g. everything in those modules that end in Result?

BurningAXE: In general yes. However, I wonder if it's worth the effort!? As the code is open now, there's no benefit to obfuscation besides a very small reduction in binary size. Given the amount of bugs that obfuscation causes I'd even argue that disabling obfuscation for the whole project may be a good idea.

PR #584, opened 2024-02-28
MS-157 Login form updates

luhmirin-s:
Main changes: 
* Removed unnecessary "hide password" toggle for user ID field
* Made project ID field read-only
* Made disabled input fields more obviously disabled
* Moved QR project ID comparison to the one provided by intent to happen right after the scan
* Added an option to manually change the backend URL for cases when QR code cannot be used for the login (extreme edge case, but still a valid one)

### DIFF
--- a/infra/network/src/test/java/com/simprints/infra/network/url/BaseUrlProviderImplTest.kt
+++ b/infra/network/src/test/java/com/simprints/infra/network/url/BaseUrlProviderImplTest.kt
@@ -14,7 +14,10 @@ import org.junit.Test
 class BaseUrlProviderImplTest {
 
     companion object {
+        private const val URL_SUFFIX = "/androidapi/v2/"
+
         private const val URL = "https://test"
+        private const val URL_WITH_SUFFIX = "https://test$URL_SUFFIX"
     }
 
     @RelaxedMockK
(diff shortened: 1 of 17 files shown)


alexandr-simprints (approved):

infra/network/src/test/java/com/simprints/infra/network/url/BaseUrlProviderImplTest.kt
```diff
@@ -14,7 +14,10 @@ import org.junit.Test
 class BaseUrlProviderImplTest {
 
     companion object {
+        private const val URL_SUFFIX = "/androidapi/v2/"
+
         private const val URL = "https://test"
+        private const val URL_WITH_PREFIX = "https://test$URL_SUFFIX"
```
alex-vt, line 20: Was the intended name `URL_WITH_SUFFIX`?

alex-vt (approved):

infra/network/src/test/java/com/simprints/infra/network/url/BaseUrlProviderImplTest.kt
luhmirin-s: Yes, fixed.

PR #591, opened 2024-03-04
MS-223 'Scanner not connected' screen saves state after the rotation

alexandr-simprints:


luhmirin-s (approved):

fingerprint/connect/src/main/java/com/simprints/fingerprint/connect/screens/controller/ConnectScannerControllerFragment.kt
```diff
@@ -48,7 +49,8 @@ internal class ConnectScannerControllerFragment : Fragment(R.layout.fragment_con
     private var shouldRequestPermissions = true
 
     private val args: ConnectScannerControllerFragmentArgs by navArgs()
-    private val viewModel: ConnectScannerViewModel by activityViewModels()
+    private val activityViewModel: ConnectScannerViewModel by activityViewModels()
+    private val fragmentViewModel: ConnectScannerControllerViewModel by viewModels()
```
meladRaouf, line 53: Is there a way to simplify this structure by eliminating one of the view models? The current functionality of ConnectScannerControllerViewModel might be directly integrated into ConnectScannerViewModel. This could also address the potential naming ambiguity.

meladRaouf (changes requested):

fingerprint/connect/src/main/java/com/simprints/fingerprint/connect/screens/controller/ConnectScannerControllerFragment.kt
alexandr-simprints: The issue is that `activityViewModel` is received by the `activityViewModels()`, meaning that this viewmodel's lifecycle is attached to the hosting activity (OrchestratorActivity). The latter should not know, nor care about the lifecycle of the initialization of the underlying fragments.

We need to make sure that the initialization of the `ConnectScannerControllerFragment` happens only once throughout its lifecycle and that's why `fragmentViewModel` is added. It represents the view model of the fragment (_not its activity_) and therefore, should hold the initialization state.

meladRaouf (approved):

fingerprint/connect/src/main/java/com/simprints/fingerprint/connect/screens/controller/ConnectScannerControllerFragment.kt
```diff
@@ -84,17 +86,19 @@ internal class ConnectScannerControllerFragment : Fragment(R.layout.fragment_con
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        val isFirstLaunch = !fragmentViewModel.isInitialized
         shouldRequestPermissions = savedInstanceState?.getBoolean(KEY_SHOULD_REQUEST_PERMISSIONS)
             ?: shouldRequestPermissions
 
-        if (savedInstanceState == null) {
-            viewModel.init(args.params)
+        if (isFirstLaunch) {
```
BurningAXE, line 93: Maybe we can get rid of this var and use the VMs one directly.

BurningAXE (approved):

PR #598, opened 2024-03-05
Add a guard to prevent parallel flow execution

luhmirin-s:
Do not start any orchestration flows if there is one already executing.

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
@@ -5,36 +5,52 @@ import android.os.Bundle
 import androidx.core.os.bundleOf
 import androidx.navigation.findNavController
 import com.simprints.core.tools.activity.BaseActivity
+import com.simprints.feature.orchestrator.cache.OrchestratorCache
 import com.simprints.feature.orchestrator.databinding.ActivityOrchestratorBinding
+import com.simprints.infra.logging.Simber
 import com.simprints.infra.orchestration.data.results.AppResult
 import com.simprints.infra.uibase.navigation.handleResult
 import com.simprints.infra.uibase.viewbinding.viewBinding
 import dagger.hilt.android.AndroidEntryPoint
+import javax.inject.Inject
 
 @AndroidEntryPoint
 internal class OrchestratorActivity : BaseActivity() {
 
     private val binding by viewBinding(ActivityOrchestratorBinding::inflate)
 
+    @Inject
+    lateinit var orchestratorCache: OrchestratorCache
+
     override fun onCreate(savedInstanceState: Bundle?) {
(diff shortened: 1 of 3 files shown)


feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorFragment.kt
```diff
@@ -30,6 +30,7 @@ import com.simprints.infra.orchestration.data.responses.AppVerifyResponse
 import com.simprints.feature.selectsubject.SelectSubjectContract
 import com.simprints.feature.setup.SetupContract
 import com.simprints.fingerprint.capture.FingerprintCaptureContract
+import com.simprints.infra.logging.Simber
```
BurningAXE, line 33: Remains from debugging?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/cache/OrchestratorCache.kt
```diff
@@ -48,6 +49,16 @@ internal class OrchestratorCache @Inject constructor(
         }
     }
 
+    /**
+     * There have been a number of cases when calling up makes several requests tens of ms apart, while it
```
BurningAXE, line 53: calling app*

BurningAXE (approved):
Looks like it might do the job but we should test with all 3 major data collectors before releasing to ensure we're not causing bigger trouble.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
```diff
@@ -5,36 +5,50 @@ import android.os.Bundle
...
+        if (orchestratorCache.isExecuting.compareAndSet(false, true))  {
+            val action = intent.action.orEmpty()
+            val extras = intent.extras ?: bundleOf()
+
+            findNavController(R.id.orchestrationHost).setGraph(
+                R.navigation.graph_orchestration,
+                OrchestratorFragmentArgs(action, extras).toBundle()
+            )
+        } else {
+            finish()
```
BurningAXE, line 50: Maybe we should add a Simber.e() here to track how often this happens!?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorFragment.kt
luhmirin-s: Accident, removed.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
luhmirin-s: Yeah, I had a message in a different spot and forgot to add it here when changing the solution.

PR #599, opened 2024-03-05
Fun with flags

luhmirin-s:
* @BurningAXE, added handling of the case you asked in the previous PR
* @alexandr-simprints, added a fix for the issue you found - should have included activity life cycle tracking in the first place, tbh.

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
@@ -1,0 +1,38 @@
+package com.simprints.feature.orchestrator
+
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleEventObserver
+import androidx.lifecycle.LifecycleOwner
+import com.simprints.infra.logging.Simber
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class ExecutionTracker @Inject constructor() : LifecycleEventObserver {
+
+    /**
+     * There have been a number of cases when calling app makes several requests tens of ms apart, while it
+     * is not exactly clear why this is happening, it is clear that it is not the intended behavior.
+     *
+     * This in-memory flag is used to prevent the orchestrator from starting multiple times.
+     * It will cause any consecutive requests to return `Activity.RESULT_CANCELED` and it is callers
+     * responsibility to handle this case.
+     */
+    val isExecuting = AtomicBoolean(false)
+
+    private var activityCount = 0
+
+    override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_CREATE) {
+            activityCount++
+        } else if (event == Lifecycle.Event.ON_DESTROY) {
+            activityCount = (activityCount - 1).coerceAtLeast(0)
+            if (activityCount == 0) {
+                isExecuting.set(false)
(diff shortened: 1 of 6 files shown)


feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
```diff
@@ -0,0 +1,38 @@
...
+     * responsibility to handle this case.
+     */
+    val isExecuting = AtomicBoolean(false)
+
+    private var activityCount = 0
+
+    override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
+        if (event == Lifecycle.Event.ON_CREATE) {
+            activityCount++
+        } else if (event == Lifecycle.Event.ON_DESTROY) {
```
BurningAXE, line 29: Hmm, what was the use case that this is handling?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorFragment.kt
```diff
@@ -83,6 +81,11 @@ internal class OrchestratorFragment : Fragment(R.layout.fragment_orchestrator) {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
+        if (savedInstanceState != null) {
+            // Only updating when explicitly restoring the state
+            orchestratorVm.requestProcessed = savedInstanceState.getBoolean(KEY_REQUEST_PROCESSED)
```
BurningAXE, line 86: If the flag is moved out of the Fragment, then we don't need to save and restore it anymore?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
luhmirin-s: There was a case where you could close an activity before the result callback, and it got stuck with the flag in the "true" state. This ensures that the flag is reset when all activities are closed.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorFragment.kt
BurningAXE: Hmm, I'd argue that this makes it more complex and error-prone. Previously, the request itself (in the args) and the request status were kept inside the Fragment. So, if you just ensure that status is persisted through Fragment restoration, you ensure that the same request won't be handled twice.
Now, a central place keeps the status but if we have two Fragment instances which request is the status referring to?

luhmirin-s: Afaik, the view model is not shared between fragment instances in the case of the parallel flow. Although due to the other fix it should not be an issue anymore.

luhmirin-s: After some poking around, this helps when restoring the screen after minimisation. I would rather keep the flag stored in a saved state for now. At least until we figure out the correct way to restore the execution flow.

meladRaouf (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorFragment.kt
BurningAXE: Oh, right! FSR I was thinking of the Orchestrator as a singleton while yeah, it's an instance connected to the particular Activity/Fragment pair.

PR #604, opened 2024-03-07
[MS-255] Fixing and issue when the biometrics flow is not restored after moving application to the background

alexandr-simprints:
ExecutionTracker now saves the 'currentLifecycleOwnerId' as a reference to the current main executing activity. All subsequent OrchestratorActivity instances will be finished, if another one is executing.

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
@@ -3,36 +3,70 @@ package com.simprints.feature.orchestrator
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
+import com.simprints.core.tools.time.TimeHelper
 import com.simprints.infra.logging.Simber
-import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-internal class ExecutionTracker @Inject constructor() : LifecycleEventObserver {
+internal class ExecutionTracker @Inject constructor(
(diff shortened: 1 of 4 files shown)


feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
```diff
@@ -3,36 +3,70 @@ package com.simprints.feature.orchestrator
...
+    /**
+     * Check for whether enough time has passed since the last set of the [currentLifecycleOwnerId].
+     *
+     * @return true when the difference between current time and the [timestamp] is greater than
+     * amount of seconds specified in the [executionTimeLimitSec]. False otherwise.
+     */
+    private fun enoughTimePassedSinceLastLock() =
+        timeHelper.msBetweenNowAndTime(timestamp) > 1000 * executionTimeLimitSec
+
+    fun LifecycleOwner.id(): Int = hashCode()
```
luhmirin-s, line 59: Why not use hashCode() directly?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/ExecutionTracker.kt
```diff
@@ -3,36 +3,70 @@ package com.simprints.feature.orchestrator
...
 import com.simprints.infra.logging.Simber
-import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-internal class ExecutionTracker @Inject constructor() : LifecycleEventObserver {
+internal class ExecutionTracker @Inject constructor(
+    private val timeHelper: TimeHelper,
+    @ExecutorLockTimeoutSec private val executionTimeLimitSec: Int
```
luhmirin-s, line 14: This seems like a bit of overengineering, it could have been a simple `const val` in this class, IMO.

PR #605, opened 2024-03-10
[MS-256] filter out unsupported templates while matching

meladRaouf:


alex-vt (approved):

fingerprint/infra/nec-bio-sdk/src/main/java/com/simprints/fingerprint/infra/necsdkimpl/matching/FingerprintMatcherImpl.kt
```diff
@@ -19,10 +19,18 @@ internal class FingerprintMatcherImpl @Inject constructor(
         candidates: List<FingerprintIdentity>,
         settings: NecMatchingSettings?
     ): List<MatchResult> {
+        // if probe template format is not supported by NEC matcher, return empty list
+        if (probe.templateFormatNotSupportedByNecMatcher()) {
+            return emptyList()
+        }
+        // if any candidate template format is not supported by NEC matcher, ignore those candidates
+        val supportedCandidates =
```
luhmirin-s, line 27: This looks like an extra list traversal. How bad is the effect on performance when large modules are matched?

luhmirin-s: Depending on the size of the module, it could be significant.

meladRaouf: @BurningAXE Yes, this is a nice idea. I have just implemented it only to read supported templates.
Its only downside is to let the [FingerprintMatcherUseCase.kt](https://github.com/Simprints/Android-Simprints-ID/pull/605/files#diff-ec9bbacf481ebdebc8b6928eb466dc30167ece0dd5cd7042597fba1f84249119) know about which matcher is being used which means we should update it once we integrate a new matcher.
the old approach was agnostic

luhmirin-s (approved):

PR #618, opened 2024-03-13
Add event version to top-level in ApiEvent request body

luhmirin-s:


ybourgery:
@luhmirin-s With this fix you will still send the version in the payload itself, no?

ybourgery (changes requested):
@luhmirin-s  With this fix you will still send the version in the payload itself which is a bit of a waste bandwidth.

luhmirin-s:
@ybourgery updated code

ybourgery (approved):

PR #629, opened 2024-03-19
MS-277 Ensure that missing finger splash is closed even if it was paused

luhmirin-s:
* If the app was paused while the missing finger screen is shown (not sure why that would happen, tho) calling dismiss would crash.
* Moved the delay logic to VM so that it survives restoration.

### DIFF
--- a/fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/views/tryagainsplash/TryAnotherFingerSplashDialogFragment.kt
+++ b/fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/views/tryagainsplash/TryAnotherFingerSplashDialogFragment.kt
@@ -9,9 +9,9 @@ import dagger.hilt.android.AndroidEntryPoint
 import com.simprints.infra.resources.R as IDR
 
 @AndroidEntryPoint
-internal class FullScreenSplashDialog : DialogFragment(R.layout.activity_splash_screen) {
+internal class TryAnotherFingerSplashDialogFragment : DialogFragment(R.layout.activity_splash_screen) {
 
-    private val vm: FullScreenSplashViewModel by viewModels()
+    private val vm: TryAnotherFingerViewModel by viewModels()
 
     override fun getTheme() = IDR.style.Theme_Simprints_Dialog_FullScreen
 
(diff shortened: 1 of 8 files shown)


meladRaouf (approved):

alexandr-simprints (commented):
I'm in favor of adding the ViewModel here. Reason: delaying a fragment is part of a business logic, which must be located within the ViewModel scope.

The issue here is entirely related to the fragment's lifecycle, and a ViewModel + single LiveData event resolves it.

If the delay was moved to the ViewModel, then the ViewModel can emit a closing event to the observer. LiveData allows us to avoid the manual lifecycle handling and the event will be closed in all the cases.

luhmirin-s:
@meladRaouf are we not excluding dialog fragments from the report as we do with the regular fragments?

meladRaouf:
> @meladRaouf, are we not excluding dialog fragments from the report as we do with the regular fragments?

@luhmirin-s Rename the class to any *Fragment or *Activity, and it would be considered as a UI class and would be excluded from the coverage reports

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/views/tryagainsplash/TryAnotherFingerSplashDialogFragment.kt
```diff
@@ -3,23 +3,25 @@ package com.simprints.fingerprint.capture.views.tryagainsplash
...
     override fun getTheme() = IDR.style.Theme_Simprints_Dialog_FullScreen
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        lifecycleScope.launch {
-            delay(TRY_DIFFERENT_FINGER_SPLASH_DELAY)
-            dismiss()
+        vm.dismiss.observe(viewLifecycleOwner) {
+            if (it) dismissAllowingStateLoss()
```
BurningAXE, line 22: A matter of taste but I'd call the boolean 'shouldDismiss' to make the logic clearer.

BurningAXE (approved):

PR #630, opened 2024-03-20
MS-2180 Remove bundle from alert results as it is now redundant

luhmirin-s:


alexandr-simprints (approved):

alexandr-simprints:
Just to provide some context behind the motivation to remove the `Bundle` field: the `Bundle` is used withing the serializable objects and the `Bundle` itself cannot be serialized. This leads to the unexpected crashes, thus it was decided to change the `Bundle` object to something that can be properly seriazlied

luhmirin-s:
Merging since the missing code coverage is configs and data classes.

PR #632, opened 2024-03-21
Ms 325 when processing more than a single navigation request the app may crash

alexandr-simprints:
One of the specifics of the navigation graph used in SID, is that the current location and the destination need to be explicitly specified. 

The navigation graph executes the navigation transactions in the queue one by one. Sometimes, when the graph receives two requests, such as `A → B` and `A → C`, the latter cannot be processed because the `B` screen is already displayed, and the `A → C` navigation is no longer possible.

Such issues are usually symptoms of poor underlying architecture, where more than a single decision mechanism is responsible for the screen’s navigation. Still, considering the amount of legacy code, it’s better not to refactor it but rather create a bandaid that won’t allow the navigation to happen if there’s already another transaction scheduled.

One way to check if there’s another transaction scheduled is to check the `NavController.destination` field. It has a className param that represents the fragment to which the last transaction was made. We should only allow the navigation to happen if the `NavController.destination` is `null` or `NavController.destination.className == currentDisplayedFragment` because if it isn’t, it means that there’s another navigation scheduled already.

Let’s see an example:

- In the initial app state, no fragment is displayed. `NavController.destination` is `null`

- Graph wants to navigate to fragment `A`. `NavController.destination` is `null`, so we’re allowing the navigation to happen.

- `NavController.destination.className` is now equal to `A`

- Fragment `A` is displayed

- Fragment `A` receives two simultaneous navigation requests:  `A → B` and `A → C`

- `A → B` navigation request is processed first. `NavController.destination.className` is `A` , and the current fragment name is `A`. Allowing the navigation to proceed.

- Now the `NavController.destination.className` becomes `B`

- `A → C` navigation request is now being processed. `NavController.destination.className` is `B`, and the current fragment name is still `A` because the transaction hasn’t had time to happen.  Navigation request is ignored because `NavController.destination.className != currentFragment`


luhmirin-s (approved):

luhmirin-s:
@alexandr-simprints Given that the rest of the quality checks are A and the "uncovered" lines are clearly a sonar issue, I suggest just merging it as-is.

BurningAXE (approved):

PR #642, opened 2024-03-27
Rename license check event file

meladRaouf:


BurningAXE:
Shouldn't we fix this on the release branch and then merge it back to main?

meladRaouf:
> Shouldn't we fix this on the release branch and then merge it back to the main?

It isn't a release blocker.
I want to limit release branch changes to bug fixing.

luhmirin-s (approved):

luhmirin-s:
I think we can just merge everything and reset the release branch. We have not event started testing anything, so there is no harm in it.

PR #644, opened 2024-03-28
Add replace attribute to force  WAKE_LOCK permission

meladRaouf:


luhmirin-s (approved):

id/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
-    <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <uses-permission android:name="android.permission.WAKE_LOCK"  tools:replace="true" />
```
BurningAXE, line 10: Shouldn't this be `tools:replace="android:name"`

alex-vt: Could be `tools:node="replace"`?

meladRaouf: In my latest commit, I removed the maxSdkVersion property that is causing the problem

BurningAXE (approved):

PR #645, opened 2024-03-28
MS-339 Unauthorised session fix

luhmirin-s:
Deleting sessions that were closed without setting the correct project ID. This happens when a session that starts before login is closed without successfully logging in.

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/EventRepositoryImpl.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/EventRepositoryImpl.kt
@@ -86,6 +86,12 @@ internal open class EventRepositoryImpl @Inject constructor(
         eventLocalDataSource.loadEventScope(downSyncEventScopeId)
 
     override suspend fun closeEventScope(eventScope: EventScope, reason: EventScopeEndCause?) {
+        if (eventScope.projectId == PROJECT_ID_FOR_NOT_SIGNED_IN) {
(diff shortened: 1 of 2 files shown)


infra/events/src/main/java/com/simprints/infra/events/EventRepositoryImpl.kt
```diff
@@ -86,6 +86,12 @@ internal open class EventRepositoryImpl @Inject constructor(
         eventLocalDataSource.loadEventScope(downSyncEventScopeId)
 
     override suspend fun closeEventScope(eventScope: EventScope, reason: EventScopeEndCause?) {
+        if (eventScope.projectId == PROJECT_ID_FOR_NOT_SIGNED_IN) {
```
BurningAXE, line 89: We don't care about any events while not logged in?

ybourgery: Only the events that are of a failed login sessions, see the Slack thread [here](https://simprints.slack.com/archives/C01BNE51M1T/p1711621715982659)

luhmirin-s: As per the thread - for now, we should only preserve the events in sessions with a valid project id.

alexandr-simprints (approved):

BurningAXE (approved):

meladRaouf (approved):

PR #654, opened 2024-04-02
RF-427 User ID fix cherry-pick 

luhmirin-s:
Same as #653 but for the next release.


meladRaouf:
@luhmirin-s Shouldn't this PR target the main branch, as we will reset the release branch before starting the testing?

luhmirin-s:
@meladRaouf I was not aware of it and have done several fixes to the current release, so it would need to be merged back into main anyway.

PR #656, opened 2024-04-04
[MS-366] Early logging of DB corruptions

alex-vt:
Provides more log info when DB corruption exception handling fails. Not a fix.


BurningAXE (changes requested):
The logs should be kept at step 3 so the Simber.i() from step 2 is recorded in the breadcrumbs.

alex-vt:
Closing because this solution is no longer needed.

PR #675, opened 2024-04-15
Translation cherry-picking

luhmirin-s:
This branch was made to keep track of translation commits applied to the main from Transifex after the release branch has been officially started. It should be merged only after all the operational languages are marked 100% verified in Transifex.


mharvey83 (approved):
Translations approved

PR #677, opened 2024-04-15
[MS-405] Saving the Action Request in the bundle so it can survive the process death

alexandr-simprints:
Initially, when the activity's process was killed (i.e. under load or with `Don't keep activities` flag turned on), the `ActionRequest` wasn't restored because the ViewModel was re-created.

This fix addresses this issue by saving the `ActionRequest` JSON in the `onSaveInstanceState`


meladRaouf:
@alexandr-simprints Should you target main or release branch?

BurningAXE:
Interesting! Are these the steps to reproduce:
1. Turn "Don't keep activities" on
2. Start a workflow
3. Send SID to background
4. Bring SID to foreground

alexandr-simprints:
Closing to target another branch

PR #679, opened 2024-04-16
[MS-413] sign out after stopping the background sync

meladRaouf:


feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
```diff
@@ -14,13 +13,12 @@ internal class LogoutUseCase @Inject constructor(
     @ExternalScope private val externalScope: CoroutineScope,
 ) {
 
-    suspend operator fun invoke() {
-        authManager.signOut()
-
+    suspend operator fun invoke() = externalScope.launch {
         // Cancel all background sync
-        externalScope.launch {
```
luhmirin-s, line 21: This will return the exact race condition that I fixed in the last PR. `authManager.signOut()` must be called in the caller's scope (view model in this case) to prevent navigation before the auth store is cleared out.

luhmirin-s (changes requested):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
luhmirin-s: Need to check if the external scope is required there, the cancellation might be fast enough to do synchronously as well.

BurningAXE: Maybe we need a comment about that to prevent other people from modifying this in the future.

meladRaouf: @luhmirin-s 
I updated this fun to Deferred 
now, the callers to the  logout can choose to wait until it finishes

@BurningAXE I added a comment

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/syncdecline/LogoutSyncDeclineFragment.kt
```diff
@@ -62,7 +64,9 @@ class LogoutSyncDeclineFragment : Fragment(R.layout.fragment_logout_sync_decline
     }
 
     private fun processLogoutConfirmation() {
-        viewModel.logout()
-        findNavController().navigate(R.id.action_logoutSyncDeclineFragment_to_requestLoginFragment)
+        lifecycleScope.launch {
+            viewModel.logout().await()
```
luhmirin-s, line 68: This feels a bit more complicated than it should. Couldn't the same be done with regular suspend functions and view model scope? Maybe with `runBlocking{}` to ensure it does wait for the execution?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
BurningAXE: Why wouldn't the caller want to wait for the operation to finish? IMO it runs the risk that somebody doesn't wait and introduces a bug this way.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/syncdecline/LogoutSyncDeclineFragment.kt
meladRaouf: I already thought about using the  runblocking as it is simpler than using Deferred
But as you can see in the screenshot runblocking inside suspended methods is useless

image

meladRaouf: Code blocks like the following would still lead to a race condition. 
```
logoutButton.setOnClickListener{
            logoutSyncViewModel.logout()
            findNavController().navigate(R.id.action_logoutSyncFragment_to_requestLoginFragment)
        }

```
------LogoutSyncViewModel-------------


```   fun logout() {
        viewModelScope.launch { logoutUseCase() }
    }
```
----------------LogoutUseCase-----------------
```suspend operator fun invoke() = runBlocking {
        // Cancel all background sync
        syncOrchestrator.cancelBackgroundWork()
        syncOrchestrator.deleteEventSyncInfo()
        // sign out the user
        authManager.signOut()
    }```

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
meladRaouf: @BurningAXE Agreed. Waiting for the logout operation to complete is crucial in all scenarios. The lint warning serves as a safeguard, preventing callers from unintentionally ignoring the Deferred result.

luhmirin-s (approved):

BurningAXE (approved):

PR #683, opened 2024-04-17
[MS-418] Don't include projectConfigurationUpdatedAt if it's empty

BurningAXE:
More info in the ticket. Opting for the much simpler solution to avoid more work and potential problems arising from changing the type of the field.

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
@@ -1,6 +1,8 @@
 package com.simprints.infra.eventsync.event.remote.models.session
 
 import androidx.annotation.Keep
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.annotation.JsonInclude.Include
 import com.simprints.infra.events.event.domain.models.Event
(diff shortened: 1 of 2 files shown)


luhmirin-s (approved):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
```diff
@@ -22,6 +24,7 @@ internal data class ApiEventScope(
     val device: ApiDevice,
     val databaseInfo: ApiDatabaseInfo,
     val location: ApiLocation?,
+    @JsonInclude(Include.NON_EMPTY)
```
ybourgery, line 27: I think you could add a test to ensure that you have what you want, and if people remove the annotation something will fail

ybourgery (changes requested):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
BurningAXE: Good idea! Added a test.

BurningAXE: Good idea! Added a test.

ybourgery (approved):

PR #691, opened 2024-04-22
MS-427 Add request ID to response headers locally

luhmirin-s:


alexandr-simprints (approved):

infra/network/src/main/java/com/simprints/infra/network/httpclient/DefaultOkHttpClientBuilder.kt
```diff
@@ -70,10 +70,12 @@ internal class DefaultOkHttpClientBuilder @Inject constructor(
...
-        val newRequest = chain.request().newBuilder()
-            .addHeader(REQUEST_ID_HEADER, UUID.randomUUID().toString())
-            .build()
-        return@Interceptor chain.proceed(newRequest)
+        val requestId = UUID.randomUUID().toString()
+
+        // Adding same header to both request and response to
+        // track the request across network logs and analytics events.
+        val newRequest = chain.request().newBuilder().addHeader(REQUEST_ID_HEADER, requestId).build()
+        return@Interceptor chain.proceed(newRequest).newBuilder().addHeader(REQUEST_ID_HEADER, requestId).build()
```
ethiery, line 78: Smart!

luhmirin-s: Not really, but this works. :D IMO, this is a bit of misuse, since the client should not add the response data 🤷🏻

ethiery: Agreed, but as far as hacks go, this is probably the smallest possible change you could make to make this work 😂 Code golf winner

PR #700, opened 2024-04-24
[MS-368] Opting out of setting worker service foreground when battery optimization on

alex-vt:
A fix for some cases of uncaught `BackgroundServiceStartNotAllowedException`, before an improved solution is found.

### DIFF
--- a/infra/core/src/main/java/com/simprints/core/tools/utils/BatteryOptimizationUtils.kt
+++ b/infra/core/src/main/java/com/simprints/core/tools/utils/BatteryOptimizationUtils.kt
@@ -2,7 +2,9 @@ package com.simprints.core.tools.utils
 
 import android.content.Context
 import android.os.PowerManager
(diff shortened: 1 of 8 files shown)


meladRaouf (approved):

infra/core/src/main/java/com/simprints/core/tools/utils/BatteryOptimizationUtils.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import android.content.Context
+import android.os.PowerManager
+
+object BatteryOptimizationUtils {
+
+    fun isFollowingBatteryOptimizations(context: Context): Boolean =
+        (context.getSystemService(Context.POWER_SERVICE) as PowerManager)
+            .isIgnoringBatteryOptimizations(context.packageName)
+            .not()
```
luhmirin-s, line 11: This reminds me of very old Borat sketch :D

PR #708, opened 2024-05-07
MS-438 Prevent invalid person creation event

luhmirin-s:
For historical reasons, the "PersonCreated" event data was collected as a combination of capture event IDs from the events in the database and captured templates in orchestrator step results.

In some erroneous edge cases, the capture data in step results and the database events might not match. Most likely, it was when one session was abruptly stopped, and another took its place (but that has to be proven). In any case, such mismatch causes invalid "PersonCreation" - one or more modalities do not have matching sets of capture reference IDs and event IDs.

This change ensures that the "CreatePerson" event has the correct sets of IDs since all the info comes only from step results. This does not address the edge case itself, but it at least makes it easier to recover the data. 

**Note**: IMO, the change, while relatively small, can have some unintended effects. Therefore, depending on the urgency, it should be considered only as part of the next hotfix or major release cycle.

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCase.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCase.kt
@@ -5,20 +5,16 @@ import com.simprints.core.domain.face.uniqueId
 import com.simprints.core.domain.fingerprint.FingerprintSample
 import com.simprints.core.domain.fingerprint.uniqueId
 import com.simprints.core.tools.time.TimeHelper
-import com.simprints.core.tools.utils.EncodingUtils
 import com.simprints.face.capture.FaceCaptureResult
 import com.simprints.fingerprint.capture.FingerprintCaptureResult
 import com.simprints.infra.events.SessionEventRepository
 import com.simprints.infra.events.event.domain.models.PersonCreationEvent
-import com.simprints.infra.events.event.domain.models.face.FaceCaptureBiometricsEvent
-import com.simprints.infra.events.event.domain.models.fingerprint.FingerprintCaptureBiometricsEvent
 import java.io.Serializable
 import javax.inject.Inject
 
 internal class CreatePersonEventUseCase @Inject constructor(
     private val eventRepository: SessionEventRepository,
     private val timeHelper: TimeHelper,
-    private val encodingUtils: EncodingUtils,
 ) {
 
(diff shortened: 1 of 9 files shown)


feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCase.kt
```diff
@@ -27,70 +23,53 @@ internal class CreatePersonEventUseCase @Inject constructor(
...
-            val fingerprintSamples = extractFingerprintSamples(results)
+            val personCreationEvent = build(faceCaptures, fingerprintCaptures)
 
-            val personCreationEvent = build(
-                faceCaptureBiometricsEvents,
-                fingerprintCaptureBiometricsEvents,
-                faceSamples,
-                fingerprintSamples
-            )
             if (personCreationEvent.hasBiometricData()) {
```
BurningAXE, line 31: Should we log an error in the opposite case?

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCaseTest.kt
```diff
@@ -70,10 +60,7 @@ internal class CreatePersonEventUseCaseTest {
 
     @Test
     fun `Does not create event if no biometric data`() = runTest {
```
BurningAXE, line 62: Do we need more tests to cover the case that is prevented with this PR?

BurningAXE (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCase.kt
luhmirin-s: We could, but I am pretty sceptical that anyone will ever look into it.

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/CreatePersonEventUseCaseTest.kt
luhmirin-s: The issue is prevented by removing the repo and only using the step result data. I think the current tests cover that; IIRC, there are some extra checks in places where the data is added to step results.

meladRaouf (approved):

alexandr-simprints (approved):

PR #710, opened 2024-05-09
[MS-437] Prevent saving of empty VM data when Fragment is destroyed and restored multiple times without being visible

BurningAXE:
To reproduce original bug:
1. Turn on "Don't keep activities"
2. Trigger a flow with dual modalities
3. Capture fingerprints
4. After face onboarding is shown minimize app
5. Bring app back to foreground
6. Capture face
7. Minimize app
8. Bring app back to foreground
9. Minimize app
10. Bring app back to foreground
11. Confirm face capture
12. Kaboom!


luhmirin-s (approved):
Those reproduction steps are getting weirder and weirder.

PR #716, opened 2024-05-14
Reverting `firebase_perf_version` from 21.0.0 to 20.5.2. 21.0.0 breaks the build due to circular dependencies

alexandr-simprints:


meladRaouf (approved):

luhmirin-s (approved):

luhmirin-s:
Should we switch to Firebase BOM and not handle each library separately?

PR #718, opened 2024-05-15
MS-338 New image saving strategy

luhmirin-s:
Main changes:
* Existing `ONLY_GOOD_SCAN` saving strategy replaced by `ONLY_USED_IN_REFERENCE` since it is how it already works.
* Adding new logic for `ONLY_GOOD_SCAN` to extract only images with quality scores over the threshold.


meladRaouf (approved):

ybourgery (changes requested):
You need to add the ONLY_USED_IN_REFERENCE for face too (it would be the same as the current implementation of ONLY_GOOD_SCANS)

luhmirin-s:
> You need to add the ONLY_USED_IN_REFERENCE for face too (it would be the same as the current implementation of ONLY_GOOD_SCANS)

🤦🏻 For unknown reason I assumed that it uses the same enum.

ybourgery (approved):

PR #726, opened 2024-05-22
[MS-474] use secugen wsq decoder

meladRaouf:
I removed the WSQ library and switched to the SecuGen WSQ library because it directly generates the needed pure image bytes. The old library was generating bitmaps, which required an extra step to convert to pure image bytes.


fingerprint/infra/nec-bio-sdk/src/test/java/com/simprints/fingerprint/infra/necsdkimpl/acquisition/template/FingerprintTemplateProviderImplTest.kt
```diff
@@ -1,5 +1,6 @@
 package com.simprints.fingerprint.infra.necsdkimpl.acquisition.template
 
+import SecuGen.FDxSDKPro.WSQConverter
```
luhmirin-s, line 3: I did not notice it in the actual project, but this namespace breaks all kinds of naming conventions. Looks almost like .Net or Objective-C import :D

luhmirin-s: Do we need to have all wrapper classes in that package, tho? Maybe we could have an explicit separation - correct package name for wrapper, forced package name for the JNI parts?

meladRaouf: Done

luhmirin-s (approved):

BurningAXE (approved):

PR #735, opened 2024-05-29
[MS-226] LiveFeedbackFragment now utilizes different layouts for horizontal and vertical screen orientations

alexandr-simprints:
Due to historical reasons, the preview is formed around the guideline in the layout, whose height is defined as percentage of the screen height. It is hardcoded to 0.3 (30% of sceren height), and it doesn't work for the smaller horizontal dimensions. Thus, the new horizontal layout is introduced, that has a guideline set at the 0.5 (50% of screen height).

face-capture-1
face-capture-2
face-capture-3
face-capture-4
image

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/models/ScreenOrientation.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/models/ScreenOrientation.kt
@@ -1,0 +1,16 @@
+package com.simprints.face.capture.models
+
+import android.content.res.Configuration
+import android.content.res.Resources
+
+enum class ScreenOrientation {
+    Landscape, Portrait;
+
+    companion object {
+        fun getCurrentOrientation(resources: Resources) =
+            when (resources.configuration.orientation) {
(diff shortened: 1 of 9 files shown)


BurningAXE (approved):

face/capture/src/main/java/com/simprints/face/capture/models/ScreenOrientation.kt
```diff
@@ -0,0 +1,16 @@
+package com.simprints.face.capture.models
+
+import android.content.res.Configuration
+import android.content.res.Resources
+
+enum class ScreenOrientation {
```
luhmirin-s, line 6: As per the latest guidelines, we should use https://developer.android.com/reference/androidx/window/core/layout/WindowSizeClass instead of the orientation.

alexandr-simprints: We only require the orientation from the resources (the layout being used by the system - regular one or the one from the `layouts/land` folder). Getting the orientation from the `resources` is exactly what is required in this case

PR #738, opened 2024-05-30
MS-219 ID pool validation

luhmirin-s:
Notable changes:
* Added a new step to the identification flow to validate the potential pool of candidates before capturing biometrics.
* If the candidate pool is empty, several conditions are checked to determine the most useful error message. (the error message wording is not final and could be updated via Transifex)
* If the data has not been synced for longer than "maxAge" configuration, the user will be presented with an option to sync and retry.


feature/validate-subject-pool/src/main/res/layout/fragment_validate_subject_pool.xml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="utf-8"?>
```
alexandr-simprints, line 1: We've previously had issues with the horizontal orientation when using the `layout_constraint*_bias` and `layout_constraint*_percent` params. Can you validate that this layout looks properly when the mobile device is rotated?

infra/resources/src/main/res/values/strings.xml
```diff
@@ -424,4 +424,15 @@
     <!-- Notifications -->
     <string name="notification_sync_description">Sync in progress</string>
 
+    <!-- Identification pool validation -->
```
alexandr-simprints, line 427: Are these strings added to the translation service?

infra/config-store/src/main/java/com/simprints/infra/config/store/local/migrations/models/OldProjectConfig.kt
```diff
@@ -224,7 +224,8 @@ internal data class OldProjectConfig(
                     if (syncGroup == "GLOBAL") "PROJECT" else syncGroup
                 ),
                 maxNbOfModules = maxNbOfModules.toInt(),
-                moduleOptions = moduleIdOptions.split("|").map(String::asTokenizableRaw)
+                moduleOptions = moduleIdOptions.split("|").map(String::asTokenizableRaw),
+                maxAge = "PT24H"
```
alexandr-simprints, line 228: Shouldn't the constant be used instead of `PT24H`?

alexandr-simprints (approved):
Approving overall, but please address some of the comments

feature/validate-subject-pool/src/main/res/layout/fragment_validate_subject_pool.xml
luhmirin-s: I checked both orientations in the preview. The layout is very basic so I think it should be fine.

infra/resources/src/main/res/values/strings.xml
luhmirin-s: They will be added automatically on merge.

BurningAXE:
> If the data has not been synced for longer than "maxAge" configuration, the user will be presented with an option to sync and retry.
Can this be named something other than age to avoid confusion (and mixed search results) with subjectAge which is about to be introduced shortly.

PR #744, opened 2024-06-03
MS-421 Login error message adjusted to also make sense for an Ending project

alex-vt:
The error is the same for invalid credentials and for Ending project status, but the error message was related only to the former - adjusted.


TristramN (approved):

meladRaouf (approved):

alexandr-simprints (approved):

luhmirin-s:
Could this be done directly in Transifex?

alex-vt:
> Could this be done directly in Transifex?

For English let's keep the change here, for consistency.

PR #747, opened 2024-06-05
MS-199 support different package names

luhmirin-s:
Notable changes:
* Adding the caller app package name to the action metadata in the orchestration flows.
* Based on the caller package name adjust the exact permission name and content provider URIs to reflect the exact caller package name.

### DIFF
--- a/feature/client-api/src/main/java/com/simprints/feature/clientapi/mappers/request/extractors/ActionRequestExtractor.kt
+++ b/feature/client-api/src/main/java/com/simprints/feature/clientapi/mappers/request/extractors/ActionRequestExtractor.kt
@@ -2,6 +2,7 @@ package com.simprints.feature.clientapi.mappers.request.extractors
 
 import android.content.Intent
 import com.simprints.feature.clientapi.extensions.extractString
+import com.simprints.feature.clientapi.models.ClientApiConstants
 import com.simprints.libsimprints.Constants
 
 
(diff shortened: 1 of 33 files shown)


feature/client-api/src/main/java/com/simprints/feature/clientapi/mappers/request/extractors/ActionRequestExtractor.kt
```diff
@@ -24,6 +25,8 @@ internal abstract class ActionRequestExtractor(private val extras: Map<String, A
     
     open fun getBiometricDataSource(): String = extras.extractString(Constants.SIMPRINTS_BIOMETRIC_DATA_SOURCE)
 
+    open fun getCallerPackageName(): String = extras.extractString(ClientApiConstants.CALLER_PACKAGE_NAME)
+
```
BurningAXE, line 29: Does this mean we expect the caller to send an extra "callerPackageName"?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
```diff
@@ -47,7 +50,7 @@ internal class DebugFragment : Fragment(R.layout.fragment_debug) {
     lateinit var eventRepository: EventRepository
 
     @Inject
-    lateinit var enrolmentRecordRepository: EnrolmentRecordRepository
+    lateinit var enrolmentRecordRepository: EnrolmentRecordLocalDataSource
```
BurningAXE, line 53: Not that it matters but why? :)

feature/client-api/src/main/java/com/simprints/feature/clientapi/mappers/request/extractors/ActionRequestExtractor.kt
BurningAXE: Oh, I saw the OrchestratorActivity - interesting hack :)

feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
luhmirin-s: The local data source does not need the "BiometricDataSource" arg :D It is resolved to the local source within the repo anyway.

PR #756, opened 2024-06-15
Add proguard rules to fix down sync

meladRaouf:
Retrofit added rules to support for R8 full mode in version 2.10, which is why we didn't encounter any issues while testing version 2024.1.0. We downgraded to version 2.9 to resolve the Jackson issue, so we had to manually add the rules.


BurningAXE (approved):

meladRaouf:
More details [here](https://github.com/square/retrofit/pull/3910/files/9831d5c3f9d28c61d29078d89645bce8ea258366#diff-50d428633d98e235831f5f9b75d7aa48897d5edc2bed30c55c9bb9ec20b36f82)

PR #774, opened 2024-06-26
SDK selection based on subject's age

BurningAXE:
These are the core changes to allow dynamic SDK selection based on age restriction of the SDKs and age of the subject.

Notable parts:
- Use the new project configuration which includes `allowedAgeRange` for each SDK
- Parse `subjectAge` from Intent metadata (if present)
- Update StepBuilder to add capture and match steps for appropriate SDKs if they are age restricted and `subjectAge` is present
- Update StepBuilder to not add capture and match steps for appropriate SDKs if they are age restricted and `subjectAge` is not present
- Update StepBuilder to add age group selection step if project is age restricted but `subjectAge` is not present
- Update Orchestrator to add appropriate capture and match steps after age group selection step
- Send 'age group not supported' error if no SDK supports the subject's age
- Update all components that relied on `bioSdkConfiguration`

Tried to keep commits thematic so the PR can be reviewed commit by commit if you want to follow the logic.


luhmirin-s (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
```diff
@@ -230,4 +337,51 @@ internal class BuildStepsUseCase @Inject constructor(
...
+        } else {
+            ageGroup?.let {
+                if (projectConfiguration.fingerprint?.secugenSimMatcher?.allowedAgeRange?.contains(ageGroup) == true) {
+                    sdks.add(FingerprintConfiguration.BioSdk.SECUGEN_SIM_MATCHER)
+                }
+                if (projectConfiguration.fingerprint?.nec?.allowedAgeRange?.contains(ageGroup) == true) {
+                    sdks.add(FingerprintConfiguration.BioSdk.NEC)
+                }
+            }
+        }
```
ybourgery, line 358: I don't think we can support multiple fingerprint SDKs being used for a reference currently as for example for the `PersonCreation` event we only have one `fingerprintReferenceId`

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
```diff
@@ -103,23 +82,126 @@ internal class BuildStepsUseCase @Inject constructor(
         )
     }.flatten()
 
-    private fun buildAgeSelectionStep(
+    fun buildCaptureAndMatchStepsForAgeGroup(
```
ybourgery, line 85: I think this logic is complex enough that you can add unit tests for it to ensure the correct SDKs are selected

ybourgery (changes requested):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/fingerselection/FingerSelectionViewModel.kt
```diff
@@ -15,16 +15,20 @@ internal class FingerSelectionViewModel @Inject constructor(
...
 
     fun start() {
         viewModelScope.launch {
-            _fingerSelections.postValue(
-                configManager.getProjectConfiguration().fingerprint!!.bioSdkConfiguration.fingersToCapture
-                    .toFingerSelectionItems()
-            )
+            val fingerSelections = mutableListOf<FingerSelectionSection>()
+            configManager.getProjectConfiguration().fingerprint?.secugenSimMatcher?.fingersToCapture?.let {
+                fingerSelections.add(FingerSelectionSection("SimMatcher", it.toFingerSelectionItems()))
```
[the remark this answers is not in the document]
BurningAXE: We don't have anything like a name parameter for SDK config. I can put the names in the resources so they can be reused but this is the first time they appear in UI.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorViewModel.kt
```diff
@@ -168,7 +192,13 @@ internal class OrchestratorViewModel @Inject constructor(
             }
         }
         if (currentStep.id == StepId.FINGERPRINT_CAPTURE && result is FingerprintCaptureResult) {
-            val matchingStep = steps.firstOrNull { it.id == StepId.FINGERPRINT_MATCHER }
+            val captureParams = currentStep.payload.getParcelable<FingerprintCaptureParams>("params")
+            // Find the matching step for the same fingerprint SDK as there may be multiple match steps
+            val matchingStep = steps.firstOrNull { step ->
+                (step.id == StepId.FINGERPRINT_MATCHER)
+                        && (step.payload.getParcelable<MatchStepStubPayload>(MatchStepStubPayload.STUB_KEY)?.fingerprintSDK
+                        == captureParams?.fingerprintSDK)
```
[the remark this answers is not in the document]
BurningAXE: The `==` doesn't bother me at all but refactoring does help with readability

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
```diff
@@ -103,23 +82,126 @@ internal class BuildStepsUseCase @Inject constructor(
...
-        )
+        ).flatten()
+    }
 
+    private fun buildModalityCaptureAndMatchStepsForVerify(
+        action: ActionRequest.VerifyActionRequest,
+        projectConfiguration: ProjectConfiguration,
+        ageGroup: AgeGroup? = null,
+    ): List<Step> {
+        val resolvedAgeGroup = ageGroup ?: ageGroupFromSubjectAge(action, projectConfiguration)
```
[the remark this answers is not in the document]
BurningAXE: The problem is that all 3 functions with repeated code can be called from 2 places (which is why they were extracted in the first place). Was thinking of extracting this part in a helper function but IMO it would hardly improve things - less repetition but more lines of code.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
BurningAXE: Definitely! This has become the heart of SDK selection so I was already doing this.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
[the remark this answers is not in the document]
BurningAXE: That said - if you can think of a good solution - it's more than welcome!

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
BurningAXE: I think this should be left to the configuration. Long term we are moving in the direction of supporting multiple SDKs in the same workflow. Also, there are discussion about removing the age selection screen which might force us to capture and search with multiple SDKs during identification for example. 
All in all to say - I prefer to have flexible code from the beginning so we can easily build upon it in the future.

ybourgery (approved):

ybourgery (changes requested):
Something that I don't think was discussed in the design document, but with the current implementation the field `modalities` in the Session is based on the configuration. I think with this work it should be based on the configuration and the actual modalities that we are capturing based on the age.

BurningAXE:
> Something that I don't think was discussed in the design document, but with the current implementation the field `modalities` in the Session is based on the configuration. I think with this work it should be based on the configuration and the actual modalities that we are capturing based on the age.

Hm, what is this field used for in the backend?

ybourgery:
> > Something that I don't think was discussed in the design document, but with the current implementation the field `modalities` in the Session is based on the configuration. I think with this work it should be based on the configuration and the actual modalities that we are capturing based on the age.
> 
> Hm, what is this field used for in the backend?

We use this to know what are the different modalities that are present in the sessions to compute the different metrics/models in Clio

ybourgery (approved):

infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/FaceConfiguration.kt
```diff
@@ -20,10 +35,24 @@ internal fun FaceConfiguration.ImageSavingStrategy.toProto(): ProtoFaceConfigura
...
-        imageSavingStrategy.toDomain(),
-        decisionPolicy.toDomain(),
+        allowedSDKs = allowedSdksList.map { it.toDomain() },
+        if (hasRankOne()) rankOne.toDomain() else null,
+    )
+
+@Suppress("SameReturnValue")
+internal fun ProtoFaceConfiguration.ProtoBioSdk.toDomain() = when (this) {
+    ProtoFaceConfiguration.ProtoBioSdk.RANK_ONE -> FaceConfiguration.BioSdk.RANK_ONE
+    ProtoFaceConfiguration.ProtoBioSdk.UNRECOGNIZED -> FaceConfiguration.BioSdk.RANK_ONE
```
alexandr-simprints, line 45: `ProtoFaceConfiguration.ProtoBioSdk.UNRECOGNIZED -> FaceConfiguration.BioSdk.RANK_ONE`
Just making sure that this is a correct mapping

alexandr-simprints (approved):

infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/FaceConfiguration.kt
BurningAXE: Yes, it's correct. Kotlin wants us to always have `UNRECOGNIZED` but we only have one face SDK currently so we map to it.

PR #775, opened 2024-06-27
Remove a  test since it doesn't add any value;

meladRaouf:
1. Remove a test since it doesn't add any value;
2. Refactor AgeGroup isEmpty check to include endExclusive equals zero


infra/config-store/src/test/java/com/simprints/infra/config/store/local/models/FingerprintConfigurationTest.kt
```diff
@@ -58,18 +58,6 @@ class FingerprintConfigurationTest {
             assertThat(it.value.toProto()).isEqualTo(it.key)
         }
     }
-    @Test
```
luhmirin-s, line 61: But how about the coverage :D

luhmirin-s (approved):

alexandr-simprints (approved):

infra/config-store/src/test/java/com/simprints/infra/config/store/local/models/FingerprintConfigurationTest.kt
meladRaouf: The allowed age groups was already covered in other tests

PR #784, opened 2024-07-02
Roc v3 into sid

meladRaouf:
1. New Modules for Roc v3 and bio SDK resolver
2. filter out unwanted templates from identification 
**Note for testing**
make sure to terminate the app if you want to switch between ROC V1 and V3, as only one `.so `library can be loaded at a time. This shouldn't be a problem, as users are not permitted to switch between different SDKs during use.


feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -43,7 +51,11 @@ internal class FaceMatcherUseCase @Inject constructor(
         val resultItems = createRanges(totalCandidates)
             .map { range ->
                 async(dispatcher) {
-                    val batchCandidates = getCandidates(matchParams.queryForCandidates, range, dataSource = matchParams.biometricDataSource)
+                    val batchCandidates = getCandidates(
+                        matchParams.queryForCandidates,
```
BurningAXE, line 55: I think `queryWithSupportedFormat` should also be used here

BurningAXE: Maybe also add a test to ensure this does not regress by mistake in the future.

face/infra/roc-v3/src/main/java/com/simprints/face/infra/rocv3/detection/RocV3Detector.kt
```diff
@@ -0,0 +1,165 @@
...
+        )
+        roc3.roc_bgr2gray(rocColorImage, rocGrayImage)
+        return detectFace(rocColorImage, rocGrayImage, bitmap.width, bitmap.height)
+    }
+
+    /*
+    To detect a face from image
+    1- use the gray image to detect the face
+    2- if the face is detected, then use the gray image to generate the landmarks
+    3- use the landmarks and the color image to generate the face template
```
BurningAXE, line 44: Interesting! Is this the recommended way to do things by ROC?

BurningAXE (changes requested):
Only a small mistake (I think). 
Otherwise - very good job, especially given the record time it took to implement!

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
meladRaouf: @BurningAXE Done

face/infra/roc-v3/src/main/java/com/simprints/face/infra/rocv3/detection/RocV3Detector.kt
meladRaouf: That is how they do it in the samples. But not written in the docs anywhere

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragmentViewModel.kt
```diff
@@ -38,9 +39,7 @@ internal class LiveFeedbackFragmentViewModel @Inject constructor(
     private var samplesToCapture: Int = 1
 
     private val faceTarget = FaceTarget(
-        SymmetricTarget(VALID_YAW_DELTA),
-        SymmetricTarget(VALID_ROLL_DELTA),
-        0.25f..0.5f
+        SymmetricTarget(VALID_YAW_DELTA), SymmetricTarget(VALID_ROLL_DELTA), 0.25f..0.5f
```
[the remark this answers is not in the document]
meladRaouf: It was the auto formatter

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
BurningAXE: Looks like `matchParams.queryForCandidates` is still used instead of `queryWithSupportedFormat`

meladRaouf: It appears to be a glitch in GitHub's diff view. You can check out the code to see that the change has already been made.
image

alexandr-simprints (approved):

BurningAXE (approved):

PR #792, opened 2024-07-08
[MS-560] Login screen buttons are now the same width as input fields

alexandr-simprints:
image
image


meladRaouf (approved):

jonsimprints (approved):
Only commenting on UI - looks good

PR #801, opened 2024-07-09
Read scopes to upsync in batches

BurningAXE:


infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
```diff
@@ -229,8 +237,10 @@ internal class EventUpSyncTask @Inject constructor(
...
      */
-    private suspend fun FlowCollector<Int>.getClosedScopesForType(type: EventScopeType) =
-        eventRepository.getClosedEventScopes(type).associateWith {
+    private suspend fun FlowCollector<Int>.getClosedScopesForType(
+        type: EventScopeType,
+        limit: Int,
+    ) =  eventRepository.getClosedEventScopes(type, limit).associateWith {
             try {
                 eventRepository.getEventsFromScope(it.id)
                     .also { listOfEvents -> emit(listOfEvents.size) }
```
meladRaouf, line 246: @BurningAXE here you will emit to the progress bar only the partial events count for the selected scope not all scopes

BurningAXE: Yes, that's the only drawback I could think of - it will lead to jumpy progress bar, too.

meladRaouf: @BurningAXE Overall, well done in such a short time.

PR #804, opened 2024-07-10
Read scopes to upsync in batches

BurningAXE:


infra/events/src/main/java/com/simprints/infra/events/EventRepository.kt
```diff
@@ -20,6 +20,8 @@ interface EventRepository {
     suspend fun saveEventScope(eventScope: EventScope)
     suspend fun getOpenEventScopes(type: EventScopeType): List<EventScope>
     suspend fun getClosedEventScopes(type: EventScopeType): List<EventScope>
```
meladRaouf, line 22: I think this is unused

meladRaouf (approved):

alexandr-simprints (approved):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
```diff
@@ -170,45 +172,48 @@ internal class EventUpSyncTask @Inject constructor(
...
         Simber.tag(SYNC_LOG_TAG)
             .d("Uploading event scope - $eventScopeTypeToUpload in batches of $batchSize")
-        val sessionScopes = getClosedScopesForType(eventScopeTypeToUpload)
 
-        // Re-emitting the number of uploaded corrupted events
-        attemptInvalidEventUpload(
-            projectId,
-            sessionScopes.getCorruptedScopes()
-        ).collect { emit(it) }
+        while (eventRepository.getClosedEventScopesCount(eventScopeTypeToUpload) > 0 && currentCoroutineContext().isActive) {
```
ybourgery, line 176: I'm not sure how much time it takes to count the closed event scopes, but I think if you use a `do while` you can avoid doing it by checking if the size of the `sessionScopes` is lower than the batch size

BurningAXE: That's how I did it initially but changed it due to concerns about error handling and creating an endless loop. @alexandr-simprints thoughts on performance vs robustness?

alexandr-simprints: > how much time it takes to count the closed event scopes
Insignificantly small amount of time, I can't see a need for an upfront optimization here

ybourgery: Ok thanks for the explanation :)

PR #805, opened 2024-07-10
[MS-572] EventDownSyncRequest/EventUpSyncRequest events error type changed to just class name

alex-vt:
The `errorType` in `EventDownSyncRequestEvent` and `EventUpSyncRequestEvent` used to contain the stack trace, which was excessive for logging at the backend.

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
@@ -75,7 +75,7 @@ internal class EventDownSyncTask @Inject constructor(
                 .catch {
                     // Track a case when event stream is closed due to a parser error,
                     // but the exception is handled gracefully and channel is closed correctly.
-                    errorType = it.toString()
+                    errorType = it.javaClass.simpleName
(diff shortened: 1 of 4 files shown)


infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
```diff
@@ -75,7 +75,7 @@ internal class EventDownSyncTask @Inject constructor(
                 .catch {
                     // Track a case when event stream is closed due to a parser error,
                     // but the exception is handled gracefully and channel is closed correctly.
-                    errorType = it.toString()
+                    errorType = it.javaClass.simpleName
```
ybourgery, line 78: Maybe you can add a test to prevent regression if someone updates this code (and same for the EventUpSyncTask)

alex-vt: @ybourgery Makes sense, added.

ybourgery (approved):

BurningAXE:
We have already released a candidate for `2024.1.4`. If this is not urgent I suggest you retarget it to `main`.

PR #807, opened 2024-07-11
[MS-577] Fetching GUID for verification only when biometricDataSource is Simprints

alex-vt:
This fixed verification with co-sync, when the biometric data source is the data collection app.
Related fix for identification: https://github.com/Simprints/Android-Simprints-ID/pull/806

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
@@ -68,7 +68,7 @@ internal class BuildStepsUseCase @Inject constructor(
         is ActionRequest.VerifyActionRequest -> listOf(
             buildSetupStep(),
             buildAgeSelectionStepIfNeeded(action, projectConfiguration),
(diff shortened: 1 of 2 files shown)


alexandr-simprints (approved):

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase.kt
```diff
@@ -222,14 +222,26 @@ internal class BuildStepsUseCase @Inject constructor(
...
-        )
-    )
+    private fun buildFetchGuidStepIfNeeded(action: ActionRequest.VerifyActionRequest) =
+        with(action) {
+            val isBiometricDataSourceSimprints =
+                BiometricDataSource.fromString(
+                    value = biometricDataSource,
+                    callerPackageName = callerPackageName,
+                ) == BiometricDataSource.Simprints
+            if (isBiometricDataSourceSimprints) {
```
BurningAXE, line 232: While there's nothing wrong with this approach - can we use the same structure that Alex used for identification so code is more consistent?

BurningAXE (approved):

PR #817, opened 2024-07-16
[MS-531] AppSweep integration

meladRaouf:
Auto-upload release APK to AppSweep server when moving build to Alpha track


mharvey83 (approved):
Looks good to me.

PR #840, opened 2024-07-30
[MS-592] Fix for face unable to be captured after app resumes from background

alex-vt:
Face capture camera viewfinder used to get stuck in its previous state after the app was put into background and then again into foreground.

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
@@ -61,6 +61,7 @@ internal class LiveFeedbackFragment : Fragment(R.layout.fragment_live_feedback) 
     private val binding by viewBinding(FragmentLiveFeedbackBinding::bind)
 
     private lateinit var screenSize: Size
+    private lateinit var targetResolution: Size
 
     private val launchPermissionRequest = registerForActivityResult(
         ActivityResultContracts.RequestPermission(),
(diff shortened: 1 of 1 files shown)


face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
```diff
@@ -104,20 +105,23 @@ internal class LiveFeedbackFragment : Fragment(R.layout.fragment_live_feedback)
...
+        if (!::targetResolution.isInitialized) {
+            targetResolution = Size(binding.captureOverlay.width, binding.captureOverlay.height)
+        }
         val imageAnalyzer = ImageAnalysis.Builder().setTargetResolution(targetResolution)
             .setOutputImageFormat(OUTPUT_IMAGE_FORMAT_RGBA_8888).build()
         imageAnalyzer.setAnalyzer(cameraExecutor, ::analyze)
         // Preview
         val preview = Preview.Builder().setTargetResolution(targetResolution).build()
         val cameraProvider = ProcessCameraProvider.getInstance(requireContext()).await()
+        cameraProvider.unbindAll()
```
BurningAXE, line 124: I don't know how CameraProvider works but since this is not obviously connected with the resolution setup, I'm just checking this is an intended change?

alex-vt: It's intended. Resolution setup is changed for just keeping the resolution and face target area the same, but this like is the part of making the capturing resume at all.

BurningAXE (approved):

luhmirin-s (approved):

meladRaouf (approved):

PR #841, opened 2024-07-30
Close session on confirm response

luhmirin-s:
Minor upkeep change that should not affect app behavior. This ensures that sessions that finish with confirmation intent have the correct end cause ("WORKFLOW_ENDED" instead of "NEW_SESSION").


ybourgery (changes requested):
What happens with existing projects that are doing several confirmations in a given session, or first a confirmation and after an enrol last biometrics? What is the reason for not considering these flows as valid?

alexandr-simprints (approved):

luhmirin-s:
> What happens with existing projects that are doing several confirmations in a given session?

@ybourgery Is that an expected behaviour? AFAIK, confirmation finishes the session. Any other flow would also close the current session and start a new one.

ybourgery:
Yes this is an expecting flow, confirmation should not close the session as you can do an enrol last biometric after a confirmation or even several confirmations. If you look at [the documentation](https://simprints.atlassian.net/wiki/spaces/KB/pages/2289172501/SID+event+documentation#Identification+-session) in Confluence, you will identification with several confirmations

PR #846, opened 2024-07-31
[MS-604] Fix SDK selection for configurations with a mix of restricted and unrestricted SDKs

BurningAXE:
Previous implementation where AllowedAgeRange for SDKs was optional had issues when the project has a mix of restricted and unrestricted SDKs. I found the cleanest way to solve this to be to make AllowedAgeRange always present. So, when an SDK is unrestricted it will have a range of [0, null) which is the same as the previous null value. This affects mostly conversion from API to domain models and domain models to/from Proto. I don't anticipate any issues but I'm a little nervous about migrations - have I missed anything?
To fix the bug I also changed the implementation of `allowedAgeRanges()` and `isAgeRestricted()` - they will now account for empty age ranges of unrestricted SDKs. I think I've covered it but this might affect any component using these functions so again - have I missed anything?


meladRaouf (approved):

meladRaouf:
@BurningAXE I have created [this test case](https://app.testiny.io/SID/testcases/tcf/33/tc/174/edit) to be executed after your PR is merged

luhmirin-s (approved):

luhmirin-s (approved):

meladRaouf (approved):

PR #847, opened 2024-08-01
[MS-612] Removing the 'fitsSystemWindows' attribute so that there is no extra white space when the toolbar is hidden

alexandr-simprints:
Once the toolbar was hidden after the `FingerprintCaptureFragment` has finished its execution, a small extra white space remained at the top of the screen.
image

This is now fixed:
image
image
image


BurningAXE (approved):
The beauty of Android flags!

luhmirin-s:
I think I have noticed something similar on exit form screen as well.

luhmirin-s (approved):

PR #878, opened 2024-08-27
MS-407 Licence renewal if face SDK init fails

luhmirin-s:
* As described in JIRA and Slack - the face capture view model now attempts to refresh the license once if the cached license is invalid for some reason.

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.simprints.core.DeviceID
 import com.simprints.core.livedata.LiveDataEvent
 import com.simprints.core.livedata.LiveDataEventWithContent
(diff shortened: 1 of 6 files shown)


face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -77,24 +83,51 @@ internal class FaceCaptureViewModel @Inject constructor(
...
         if (licenseStatus == LicenseStatus.VALID) {
-            val initializer = resolveFaceBioSdk().initializer
-            if (!initializer.tryInitWithLicense(activity, license!!.data)) {
-                // License is valid but the SDK failed to initialize
-                // This is should reported as an error
-                licenseStatus = LicenseStatus.ERROR
+            licenseStatus = initialize(activity, license!!)
+        }
+        // In some cases license is invalidated on initialisation attempt
+        if (licenseStatus != LicenseStatus.VALID) {
```
meladRaouf, line 94: @luhmirin-s It seems like there's quite a bit of code inside the `initFaceBioSdk` function. Please consider simplifying or restructuring it to make it a bit lighter.

luhmirin-s: Extracted licence refresh into a method.

alexandr-simprints (approved):

PR #891, opened 2024-09-02
MS-617 Take into account blank session IDs as well

luhmirin-s:


feature/client-api/src/test/java/com/simprints/feature/clientapi/mappers/request/IntentToActionMapperTest.kt
```diff
@@ -103,6 +103,16 @@ class IntentToActionMapperTest {
         }
     }
 
+    @Test
+    fun `correctly handles CommCare intent with blank session ID`() = runTest {
```
BurningAXE, line 107: Maybe add another test for missing sessionId as it's the real world scenario currently.

BurningAXE (approved):

feature/client-api/src/test/java/com/simprints/feature/clientapi/mappers/request/IntentToActionMapperTest.kt
luhmirin-s: It was added in the initial fix. One test case above this one.

PR #894, opened 2024-09-04
Session ID fix backport

luhmirin-s:
Cherry-picking of session ID fix/hack that was applied on 2024.1.5


BurningAXE (approved):
We need to wait for 2024.1.5 to be moved to prod before merging this.

meladRaouf (approved):

PR #895, opened 2024-09-04
[MS-710] Add UNCOOPERATIVE_CHILD exit option

meladRaouf:


luhmirin-s (approved):
Would it be an overkill to only show the option if child biometrics are being collected in the project?

meladRaouf:
> Would it be an overkill to only show the option if child biometrics are being collected in the project?
@luhmirin-s I think this is a great idea.  
@jonsimprints, what are your thoughts? Should we display this exit option only when child bio is active?

jonsimprints:
@meladRaouf how would we objectively define a child?

meladRaouf:
> @meladRaouf how would we objectively define a child?

Good question,
If any SDK is enabled and the age limit is below, for example, 6 years, we can conclude that this SDK supports child bio.

PR #905, opened 2024-09-16
[MS-491] limit fingerprint scan attempts

alexandr-simprints:
There is now a limit to the amount of 'No Finger detected' scans before the user is automatically navigated to the next finger.

- The default value is `40` and is stored in the `IsNoFingerDetectedLimitReachedUseCase.MAXIMUM_LIMIT_OF_NO_FINGER_DETECTED_SCANS`
- The value is set in the Vulcan under `Configuration -> Fingerprint -> SimMatcher`
- If the value is not in the [2..40] range, the default value of `40` is used

Other Changes
To accommodate the newly added `noFingerDetectedThreshold`, the `CaptureState` sealed class was refactored to have the following subclasses:
- `CaptureState.NotCollected`
- `CaptureState.Skipped`
- `CaptureState.ScanProcess` - number of bad scans and no-finger-detected is stored here


fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/usecase/IsNoFingerDetectedLimitReachedUseCase.kt
```diff
@@ -0,0 +1,26 @@
...
+
+internal class IsNoFingerDetectedLimitReachedUseCase @Inject constructor() {
+    operator fun invoke(
+        fingerState: CaptureState,
+        configuration: FingerprintConfiguration
+
+    ): Boolean = when (fingerState) {
+        is CaptureState.ScanProcess -> {
+            val noFingerDetectedThreshold =
+                configuration.secugenSimMatcher?.maxCaptureAttempts?.noFingerDetected?.takeIf { it > 1 }
```
BurningAXE, line 15: Is this feature only applicable to SimMatcher? If so, we should take care to either not call this use case when using NEC or pass the SDK used currently and skip the check for NEC.

BurningAXE (changes requested):
We should handle NEC, too. Otherwise looks good!

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/usecase/IsNoFingerDetectedLimitReachedUseCase.kt
alexandr-simprints: That's a good question, actually. Let's take it to Slack first

luhmirin-s (approved):

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/usecase/IsNoFingerDetectedLimitReachedUseCase.kt
BurningAXE: OK, then the use case should lookup the configuration for the particular SDK used. Currently it's always checking the SimMatcher config.

alexandr-simprints: Great spot! Fixing it now

luhmirin-s (approved):

BurningAXE (approved):

PR #906, opened 2024-09-17
MS-71 Moving image manipulation closer to CameraX API

luhmirin-s:
* Set a flag in the camera to pre-rotate images and always return them with the ROTATION_0 value
* Move cropping from the view model to the image processing pipeline in the camera API.


alexandr-simprints (commented):
Have you checked that the device rotation while in preview doesn't affect the calculations in the `CropToTargetOverlayAnalyzer`?

Otherwise, looks good

luhmirin-s:
@alexandr-simprints I did rotation during capture and it worked as expected. It seems that the camera pipeline is fully rebuilt on rotation therefore, the cropper is also rebuilt.

alexandr-simprints (approved):

PR #907, opened 2024-09-18
MS-676 Sync config button

luhmirin-s:
* Added a button to request configuration update in the Settings screen (see screenshot)
* Update is locked under the settings password (if set in Vulcan)

 Screenshot_20240918_112948


alexandr-simprints (commented):
Looks good.

Can you please try the following scenario:
- Make sure the config is password protected, so that the screen shows the dialog
- Set 'Don't keep activities' in the Developer Settings
- Navigate to `SID` -> `Settings` -> `Update Configuration`
- Make sure dialog is displayed
- Minimize the app by pressing the home button or navigating to another app
- Return to SID

Verify that only a single instance of dialog is displayed and that the dialog is functional.

Once confirmed, approving the PR

luhmirin-s:
@alexandr-simprints works as expected (it is exact same mechanism as with the language list).

alexandr-simprints (approved):

PR #913, opened 2024-09-24
Ms 720 add audio notification for finger removal after scanning

meladRaouf:


fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/FingerprintCaptureFragment.kt
```diff
@@ -104,9 +111,26 @@ internal class FingerprintCaptureFragment : Fragment(R.layout.fragment_fingerpri
...
+        mediaPlayer.start()
+        mediaPlayer.setOnCompletionListener {
+            it.release()
+        }
+    }
+
+    private fun isAudioEnabled() =
+        PreferenceManager
+            .getDefaultSharedPreferences(requireContext())
+            .getBoolean(AUDIO_PREFERENCE_KEY, true)
```
meladRaouf, line 132: I haven't yet reached an agreement with Jon regarding the default value of the preference.

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/FingerprintCaptureFragment.kt
```diff
@@ -104,9 +111,26 @@ internal class FingerprintCaptureFragment : Fragment(R.layout.fragment_fingerpri
...
 
+        scanningStatusTracker.scanCompleted.observe(viewLifecycleOwner, LiveDataEventObserver {
+            if (isAudioEnabled())
+                playBeep()
+        })
         initUI()
     }
 
+    private fun playBeep() {
+        val mediaPlayer = MediaPlayer.create(context, R.raw.beep)
```
alexandr-simprints, line 122: Can the `MediaPlayer` instance be created on the Fragment properties level so that we don't have to re-initialize it every time? If memory serves me right, the `MediaPlayer` creation might be resource-intensive due to codec and decoders initialization etc.

alexandr-simprints: If you decide that creating a media player instance makes sense, then please also make sure that it is re-created correctly once the device is rotated

fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/capture/FingerprintScanningStatusTracker.kt
alexandr-simprints, line 1: It doesn't seem to me that the `LiveData` object is the way to go in a tracker that might be used outside of the lifecycle-aware classes.

Using `SharedFlow` seems more appropriate
```
class FingerprintScanningStatusTracker @Inject constructor() {
    private val _scanCompleted = MutableSharedFlow<Unit>()
    val scanCompleted = _scanCompleted.asSharedFlow()

    suspend fun notifyScanCompleted() {
        _scanCompleted.emit(Unit)
    }
}
```

alexandr-simprints (dismissed):

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/FingerprintCaptureFragment.kt
luhmirin-s: I would also suggest moving it together with the `PreferenceManager` to some kind of injectable (ideally singleton) manager class.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Mon Apr 17 08:02:27 EET 2023
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.4-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.10-bin.zip
```
luhmirin-s, line 4: This should be a part of version bump PR, not a feature PR.

fingerprint/capture/build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
 
     // Android X
     implementation(libs.androidX.ui.viewpager2)
+    implementation(libs.androidX.ui.preference)
```
luhmirin-s, line 41: Why do you need preferences UI in this module?

luhmirin-s (approved):

gradle/wrapper/gradle-wrapper.properties
meladRaouf: I was testing the KSP rollback

fingerprint/capture/build.gradle.kts
meladRaouf: To access the PrefsManger class as it is part of the androidx prefs lib

luhmirin-s: Would it make sense to decouple the actual setting from the UI?

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/FingerprintCaptureFragment.kt
meladRaouf: The final agreement is to make it false

fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/capture/FingerprintScanningStatusTracker.kt
```diff
@@ -0,0 +1,21 @@
+package com.simprints.fingerprint.infra.scanner.capture
+
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class FingerprintScanningStatusTracker @Inject constructor() {
```
luhmirin-s, line 10: This is a bit of a nitpick, but from the name and usage in the code above, I expected this to handle multiple status values. IMO, this class should either handle all the status changes (at least have an easy way to extend to do so) or have a less generic name.

luhmirin-s (approved):

PR #918, opened 2024-09-27
[MS-742] Remove ProGuard rules from library modules and consolidate them in the app module

meladRaouf:
- Removed `proguard-rules.pro` files from all library modules.
- Consolidated ProGuard rules in the id module, as we have minimal custom rules per module.
- Simplifies project structure and ensures consistent ProGuard handling across all modules.


luhmirin-s (commented):
I see all the removals, but not many are added back to the app module. Aren't the hilt, jni and webview rules required for something?

meladRaouf:
> I see all the removals, but not many are added back to the app module. Aren't the hilt, jni and webview rules required for something?

@luhmirin-s

I found that the Hilt rules are automatically added to the app's ProGuard rules file.
 As for the other `com.simprints.**` rules, they are also included in the same app-wide file. 
About the WebView rules—where did you get them from?

luhmirin-s:
> I found that the Hilt rules are automatically added to the app's ProGuard rules file. As for the other `com.simprints.**` rules, they are also included in the same app-wide file. About the WebView rules—where did you get them from?

In this file - fingerprint/infra/simafis-wrapper/consumer-rules.pro

meladRaouf:
> > I found that the Hilt rules are automatically added to the app's ProGuard rules file. As for the other `com.simprints.**` rules, they are also included in the same app-wide file. About the WebView rules—where did you get them from?
> 
> In this file - fingerprint/infra/simafis-wrapper/consumer-rules.pro

@luhmirin-s it is a commented code

luhmirin-s (approved):

PR #924, opened 2024-10-03
MS-730 Licence version updates

luhmirin-s:
* Handling versions provided by the project configuration to proactively download the correct licence version in the setup step. 
* Additionally rearranged the public licence models into a separate package.

### DIFF
--- a/infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiFingerprintConfiguration.kt
+++ b/infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiFingerprintConfiguration.kt
@@ -31,7 +31,6 @@ internal data class ApiFingerprintConfiguration(
         val vero2: ApiVero2Configuration? = null,
         val allowedAgeRange: ApiAllowedAgeRange? = null,
         val verificationMatchThreshold: Float? = null,
-        val version: String? = null,
     ) {
         fun toDomain() = FingerprintConfiguration.FingerprintSdkConfiguration(
             fingersToCapture = fingersToCapture.map { it.toDomain() },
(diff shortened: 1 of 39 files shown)


infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiFingerprintConfiguration.kt
```diff
@@ -40,6 +41,7 @@ internal data class ApiFingerprintConfiguration(
             vero2 = vero2?.toDomain(),
             allowedAgeRange = allowedAgeRange?.toDomain() ?: AgeGroup(0, null),
             verificationMatchThreshold = verificationMatchThreshold,
+            version = version.orEmpty(),
```
ybourgery, line 44: The backend is not sending the version, so why add it into the `ApiFingerprintConfiguration` class?

luhmirin-s: For consistency and to ensure that it is handled in the same way in future.

ybourgery: IMO, for consistency, you should only add it in your domain class to make the impression to future readers that BFSID is sending this field

luhmirin-s: I will remove it for now, but I still don't think it is an issue since it is defined as optional with the default set to null, which should be a clear indicator that the version is not guaranteed in the API.

infra/config-store/src/main/proto/project_config.proto
```diff
@@ -96,6 +96,8 @@ message ProtoFingerprintConfiguration {
         optional ProtoVero1Configuration vero_1 = 5;
         optional ProtoAllowedAgeRange allowed_age_range = 6;
         optional float verification_match_threshold = 7;
+
```
meladRaouf, line 99: Extra line

meladRaouf (approved):

infra/config-store/src/main/proto/project_config.proto
luhmirin-s: Blank space is prepared for the addition of extra config that is already in main branch. Same reason it skips a number as well.

PR #931, opened 2024-10-15
MS-765 Hide ID pool validation behind a flag

luhmirin-s:
* Added a mechanism to handle dynamic custom/experimental config safely and explicitly.
* A custom configuration was added to validate ID pool only when required.

### DIFF
--- a/infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/ProjectConfiguration.kt
+++ b/infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/ProjectConfiguration.kt
@@ -1,5 +1,6 @@
 package com.simprints.infra.config.store.local.models
 
+import com.simprints.core.tools.json.JsonHelper
 import com.simprints.infra.config.store.models.ProjectConfiguration
 
 internal fun ProjectConfiguration.toProto(): ProtoProjectConfiguration =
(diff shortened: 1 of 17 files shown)


infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/ProjectConfiguration.kt
```diff
@@ -26,4 +30,5 @@ internal fun ProtoProjectConfiguration.toDomain(): ProjectConfiguration =
         consent.toDomain(),
         identification.toDomain(),
         synchronization.toDomain(),
+        customJson?.takeIf { it.isNotBlank() }?.let { JsonHelper.fromJson(it) },
```
BurningAXE, line 33: Maybe it's worth adding a catch to protect against misconfigured JSON coming from Vulcan.

infra/config-store/src/test/java/com/simprints/infra/config/store/models/ExperimentalProjectConfigurationTest.kt
```diff
@@ -0,0 +1,25 @@
+package com.simprints.infra.config.store.models
+
+import com.google.common.truth.Truth.assertThat
+import com.simprints.infra.config.store.models.ExperimentalProjectConfiguration.Companion.ENABLE_ID_POOL_VALIDATION
+import kotlin.test.Test
+
+internal class ExperimentalProjectConfigurationTest {
+
+    @Test
+    fun `check pull validation flag correctly`() {
```
BurningAXE, line 10: *pool

BurningAXE (approved):

infra/config-store/src/main/java/com/simprints/infra/config/store/local/models/ProjectConfiguration.kt
luhmirin-s: It would fail on the network communication step, so the chance is slim. But I will put it just in case of migration issues 👍🏻

PR #941, opened 2024-10-22
MS-755 Split session logging and mutex

luhmirin-s:
* Added non-fatals to see logs around session creation and restoration
* Locked every event cache interaction with mutex to prevent any potential race condition

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
@@ -6,6 +6,7 @@ import com.simprints.infra.events.event.domain.models.Event
 import com.simprints.infra.events.event.domain.models.scope.EventScope
 import com.simprints.infra.events.event.domain.models.scope.EventScopeEndCause
 import com.simprints.infra.events.event.domain.models.scope.EventScopeType
+import com.simprints.infra.logging.Simber
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import javax.inject.Inject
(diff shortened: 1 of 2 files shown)


infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
```diff
@@ -15,8 +18,11 @@ internal class SessionEventRepositoryImpl @Inject constructor(
     private val sessionDataCache: SessionDataCache,
 ) : SessionEventRepository {
 
-    override suspend fun createSession(): EventScope {
+    private val eventsLock = Mutex()
+
+    override suspend fun createSession(): EventScope = eventsLock.withLock {
         closeAllSessions(EventScopeEndCause.NEW_SESSION)
+        Simber.w("Creating session from DB")
```
BurningAXE, line 25: I think this will create too much noise

BurningAXE: Can we instead log only when we reach the call on line 54:
`?: createSession()`

BurningAXE (changes requested):
Good idea for the locks!
I have a different idea for the new session logging hence the change request.

infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
```diff
@@ -42,10 +47,19 @@ internal class SessionEventRepositoryImpl @Inject constructor(
...
+        val cachedScope = sessionDataCache.eventScope
+        if (cachedScope != null) return cachedScope
+
+        val restoredScope = localSessionScope()
+        if (restoredScope != null) {
+            Simber.w("Restored session from DB")
+            return restoredScope
+        }
+
+        Simber.w("Creating new session DB")
```
BurningAXE, line 60: Maybe "Unexpected creation of new session" or something similar :)

BurningAXE (approved):

infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
luhmirin-s: I don't think that it is unexpected in all cases.

alexandr-simprints (approved):

PR #942, opened 2024-10-22
[MS-771] Adding 'layout-sw600dp-land' layout folder specifically for the tablets.

alexandr-simprints:
The tablets are big enough to display the same layout as in the 'layout' folder, so its content is just a copy-paste of these files.

Tablet:
image
Phone:
image


luhmirin-s (approved):

meladRaouf (approved):

BurningAXE (approved):
Having two copies of the same file is not ideal but still on the border of manageable. However, if we need to do this again, I suggest we use a common layout with an <include> in the places it needs to be "duplicated".

PR #944, opened 2024-10-23
[MS-637] Matching Modalities

BurningAXE:


infra/config-store/src/main/java/com/simprints/infra/config/store/local/migrations/ProjectConfigFaceSdkQualityThresholdMigration.kt
```diff
@@ -6,13 +6,13 @@ import com.simprints.infra.logging.Simber
 import javax.inject.Inject
 
 /**
- * Can be removed once all the devices have been updated to 2022.4.1
+ * Can be removed once all the devices have been updated to 2024.2.1
```
luhmirin-s, line 9: Thanks, I goofed 🤦🏻

luhmirin-s (approved):

alexandr-simprints (approved):

infra/sync/src/test/java/com/simprints/infra/sync/config/testtools/Models.kt
```diff
@@ -20,6 +20,7 @@ import com.simprints.infra.config.store.models.Vero1Configuration
 import com.simprints.infra.config.store.models.Vero2Configuration
 
 internal val generalConfiguration = GeneralConfiguration(
+    listOf(GeneralConfiguration.Modality.FACE),
```
alexandr-simprints, line 23: Typo, duplicate?

alexandr-simprints (approved):
Approved, but double check the comment in the `Vero1Configuration` file above

infra/sync/src/test/java/com/simprints/infra/sync/config/testtools/Models.kt
BurningAXE: I suppose you are referring to the fact that you have the same line twice!?
That's correct - one is for the modalities configuration and the other is for the matchingModalities configuration. That's what happens when we don't use parameter names.

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/cache/OrchestratorCacheTest.kt
```diff
@@ -78,9 +80,31 @@ class OrchestratorCacheTest {
...
+        cache.ageGroup = AgeGroup(1, 2)
+
+        verify(exactly = 1) { prefs.edit().putString(any(), json) }
+    }
+
+    @Test
+    fun `Restores age group if stored`() {
+        val json = "[1,2]"
+        every { prefs.getString(any(), any()) } returns json
+        every { jsonHelper.fromJson<AgeGroup>(any(), any<TypeReference<AgeGroup>>()) } returns AgeGroup(1, 2)
```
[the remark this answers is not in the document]
BurningAXE: To be fair - the current unit tests are pretty moot as we mock both preferences and JSON conversion but I don't want to delay real testing for this. However, I did check that in practice (60, null) is stored and restored correctly.

PR #949, opened 2024-10-30
Fix deadlocks in session cache access

luhmirin-s:
* Ensured that locked data is not accessed from within the already locked code block by explicitly separating synchronised and unsynchronised methods.
* Added an empty method scope to explicitly mark methods that should be called from locked context.

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
@@ -20,7 +20,69 @@ internal class SessionEventRepositoryImpl @Inject constructor(
 
     private val eventsLock = Mutex()
 
-    override suspend fun createSession(): EventScope = eventsLock.withLock {
+    override suspend fun createSession(): EventScope = withLockedContext {
+        createSessionInternal()
+    }
+
+    override suspend fun saveSessionScope(eventScope: EventScope) = withLockedContext {
+        saveSessionScopeInternal(eventScope)
+    }
+
+    override suspend fun hasOpenSession(): Boolean = withLockedContext {
+        val session = sessionDataCache.eventScope ?: localSessionScope()
+        session != null
+    }
+
+    override suspend fun getCurrentSessionScope(): EventScope = withLockedContext {
+        getCurrentScopeInternal()
+    }
+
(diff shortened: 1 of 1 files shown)


alexandr-simprints (approved):

infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
```diff
@@ -20,98 +20,114 @@ internal class SessionEventRepositoryImpl @Inject constructor(
...
+    override suspend fun closeCurrentSession(reason: EventScopeEndCause?) = withLockedContext {
+        eventRepository.closeEventScope(getCurrentScopeInternal(), reason)
+        sessionDataCache.eventCache.clear()
+        sessionDataCache.eventScope = null
+    }
+
+    /**
+     * Scope for internal extension functions that must be called within locked context.
+     */
+    private object LockedContext
```
BurningAXE, line 78: Interesting approach! So, if I got this right we have the internal functions explicitly defined as extensions of this context and additionally all blocks executed with withLockedContext() are also executed in this context (even through not explicitly defined as extensions since they are lambdas)?

BurningAXE (commented):
I have two questions:
1. Did you stumble upon deadlocks with the previous solution or is it a theoretical concern (so far)?
2. How confident are you in this solution? 
LGTM and I don't see any flaws in it. However, I didn't see any flaws in the previous one, too. And this one has even more substantial changes which makes me a little nervous to deploy just ahead of the scale-up.

infra/events/src/main/java/com/simprints/infra/events/session/SessionEventRepositoryImpl.kt
luhmirin-s: Yes, internal functions are implemented as extensions to force the caller to think if the called function uses the parts that are locked with mutex. The "context" in this case does not do anything in particular besides being the receiver of the internal functions.

luhmirin-s:
@BurningAXE 
1. It is not theoretical, although the deadlock happens on an unhappy path and not all of the time, so it slipped past me initially. I was able to reproduce it consistently, tho. 
2. I am much more confident in this implementation cos it has guardrails to prevent the incorrect usage from happening again.

PR #950, opened 2024-10-31
[MS-710] Add UNCOOPERATIVE_CHILD exit option

meladRaouf:
(cherry picked from commit cf471dca0e972f163589ee188c1f2894131fb9c3)


meladRaouf:
> Does the backend support this if sent from 2024.2.1?

yes it is

PR #961, opened 2024-11-06
[MS-779] Hide capture button when permissions button is visible

meladRaouf:


face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
```diff
@@ -341,7 +349,7 @@ internal class LiveFeedbackFragment : Fragment(R.layout.fragment_live_feedback)
         binding.apply {
             captureOverlay.drawSemiTransparentTarget()
             captureFeedbackTxtExplanation.setText(IDR.string.face_capture_permission_denied)
-
+            captureFeedbackBtn.isGone = true
```
BurningAXE, line 352: It's my personal preference but I would use only `isVisible` so you can get the overall state just by looking at the booleans - requires much less cognitive load.

BurningAXE (approved):

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
meladRaouf: I totally agree with you on using `isVisible` instead of `isGone`. My only concern is that the entire file currently uses `isGone`, which is why I kept it consistent.

luhmirin-s: I would stick with `isGone` since it is cleared in the Android context since `isVisible = false` is `visibility=gone` under the hood.

PR #963, opened 2024-11-07
MS-789 singe quality fallback

luhmirin-s:
* Added a feature flag to enable more lenient capture quality check ([as described in the design doc](https://docs.google.com/document/d/1RbYeFiwQHJi4NVu3uI9syXY67rfSP2RhM5p1uAsI4no/edit?tab=t.0#heading=h.hujx7mdf5eie))

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragmentViewModel.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragmentViewModel.kt
@@ -13,6 +13,7 @@ import com.simprints.face.capture.usecases.SimpleCaptureEventReporter
 import com.simprints.face.infra.basebiosdk.detection.Face
 import com.simprints.face.infra.basebiosdk.detection.FaceDetector
 import com.simprints.face.infra.biosdkresolver.ResolveFaceBioSdkUseCase
+import com.simprints.infra.config.store.models.experimental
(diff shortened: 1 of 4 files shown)


alexandr-simprints (approved):

ybourgery (commented):
What is the reasoning for adding a feature flag vs just lowering the threshold to -1 if we see that it's not working?

BurningAXE:
Isn't this the same as setting the number of images to capture to 1?

luhmirin-s:
@ybourgery Jon had issues capturing good-quality images of his kids, so there is a chance that the user experience might get worse with the added quality check. At the same time completely disabling the quality threshold is not a great solution, since it was added for a reason. The proposed fix gives us the option to enable a more lenient capture flow if it is ever required while still preserving the quality threshold. 

@BurningAXE only if all of the actual captured images have a quality below the threshold, which might not always be the case. Also, it is the same logic we have currently, just with a guarantee that at least the fallback is of good quality.

BurningAXE:
> @BurningAXE only if all of the actual captured images have a quality below the threshold, which might not always be the case. Also, it is the same logic we have currently, just with a guarantee that at least the fallback is of good quality.
OK, so if you set the number of images to capture to 1, this option will have no effect. But if the number of images is higher and you turn it on, you are guaranteed to have at least 1 above the quality, but there's a chance to get more.

BurningAXE (approved):

PR #965, opened 2024-11-08
MS-785 Using correct context when compiling text with builders

luhmirin-s:
* When the language is changed, the Application context is not restarted and, therefore, does not reload resources. This is fine until we start using it for string building.
* This change ensures that only the activity context is used for manual string building by utilising annotations provided in Hilt and some manual factory calls.


luhmirin-s:
I have found another fix for it in the main #875

PR #966, opened 2024-11-08
MS-875 Language change fixes

luhmirin-s:
* Cherry-picked the fix for consent text resolving using activity context.
* Fixed a similar issue with age range names.


luhmirin-s:
@meladRaouf Can the duplication check be overriden? I believe it is a false positive in this case.

feature/select-subject-age-group/src/main/java/com/simprints/feature/selectagegroup/screen/BuildAgeGroupsUseCase.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.simprints.infra.config.store.ConfigRepository
+import com.simprints.infra.config.store.models.AgeGroup
+import com.simprints.infra.config.store.models.sortedUniqueAgeGroups
+import javax.inject.Inject
+
+internal class BuildAgeGroupsUseCase @Inject constructor(
+    private val configurationRepo: ConfigRepository,
+) {
+
+    /**
```
meladRaouf, line 12: Is this code comment still valid?

meladRaouf (approved):

meladRaouf:
> @meladRaouf Can the duplication check be overriden? I believe it is a false positive in this case.

We can’t override this validation, but we can still merge the PR.

feature/select-subject-age-group/src/main/java/com/simprints/feature/selectagegroup/screen/BuildAgeGroupsUseCase.kt
luhmirin-s: It is still valid, I just removed the part that was referring to display text.

PR #970, opened 2024-11-11
[MS-790] Add follow-up steps to session steps instead of replacing them

BurningAXE:
Previously each action (including follow-ups) replaced old steps (in the Orchestrator and consequently in the OrchestratorCache) with its own steps. This lead to the issue where the following actions:

1. Identify
2. Confirm Identity
3. Enrol Last
lead to Enrol Last failing because when Enrol Last looks in the Cache it only finds the step(s) for Confirm Identity.

With this fix each follow-up only adds its steps to the ones already in the cache. This way all session steps are available regardless of how many follow-ups have happened. The Cache is cleared when a new session is created so this change should not lead to mixing of steps between sessions. That said - **can you find any edge cases where this may backfire?**

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorViewModel.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorViewModel.kt
@@ -82,8 +82,11 @@ internal class OrchestratorViewModel @Inject constructor(
         actionRequest = action
 
         try {
-            steps = stepsBuilder.build(action, projectConfiguration)
-        } catch (e: SubjectAgeNotSupportedException) {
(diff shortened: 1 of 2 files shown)


alex-vt (approved):
Just in case a check: is this fix supposed to be in `release/2024.2.1`, not in `main`?

BurningAXE:
> Just in case a check: is this fix supposed to be in `release/2024.2.1`, not in `main`?

We'll decide tomorrow but there are 3 options:
1. Release in 2024.2.1 - best case scenario if considered safe
2. Have it ready for a 2024.2.2 hotfix if we don't include in 2024.2.1 but turns out to be an issue for a project
3. Just merge it in main and release in next major release (2025.1.0 I guess)

luhmirin-s (approved):

PR #974, opened 2024-11-14
MS-801 Add data export button to alert screen

luhmirin-s:
Example of data exported to the clipboard:
```
Event ID:   229155a7-1588-408a-bf92-e2f16ac6ad8c
Timestamp:  1731597053343
Project ID: xRmp384ofU6dvylF1Hao
User ID:    Sergejs
Device ID:  d294a268b0d54f58
Session ID: c36ed54b-621c-47f3-845b-87d7b52c9e7f
Alert type: INVALID_SESSION_ID
```

[Figma design](https://www.figma.com/design/8gHXXZJkAhOnb2a7bm8llk/Simprints?node-id=2421-7265&node-type=canvas&t=jbo9t3dD22cG36A9-0)

### DIFF
--- a/feature/alert/build.gradle.kts
+++ b/feature/alert/build.gradle.kts
@@ -9,4 +9,5 @@ android {
 
 dependencies {
     implementation(project(":infra:events"))
+    implementation(project(":infra:auth-store"))
 }
--- a/feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
+++ b/feature/alert/src/main/java/com/simprints/feature/alert/screen/AlertFragment.kt
@@ -3,6 +3,7 @@ package com.simprints.feature.alert.screen
 import android.os.Bundle
 import android.view.View
 import android.widget.TextView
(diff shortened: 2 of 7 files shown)


ybourgery (commented):
IMO it would be good to try to be able to export more details about the root cause of the alert screen, e.g. the exception that led to the issue as when you have an unexpected error it will help to find the root cause.

luhmirin-s:
I initially wanted to add it, but we are not propagating the actual exception to the alert screen. And in many cases, there is no exception at all. 

As a side note - I was considering logging a non-fatal with a UUID in the title or a field so that logs could be searchable in Crashlytics. But it felt a bit overkill, since we are already logging the "unexpected" exceptions directly.

ybourgery:
Ah yes but sometimes you can have a given alert for several exceptions, for example the `INVALID_SESSION_ID` can be there because the session id is empty or invalid. But if it's too complex to implement it's ok

ybourgery (approved):

PR #977, opened 2024-11-18
[MS-816] Yellow alert screen background dimmed for better text readability

alex-vt:
Brightness level for yellow Alert screen reduced by 25% to ease white text reading. This is the only use of the yellow color in the app.


luhmirin-s:
Is the new colour still within general Simprints design guides?

luhmirin-s (approved):

alex-vt:
@luhmirin-s This shouldn't be a deterioration from the design perspective: readability is improved comparing to the low-contrasting previous color, hue and saturation aren't changed, and the color appears only in a scenario of an error in the app.

PR #979, opened 2024-11-18
[MS-824] Ensure sequential execution of UN20 tasks

meladRaouf:
This PR ensures sequential execution of UN20 tasks.

### DIFF
--- a/fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/ObserveFingerprintScanStatusUseCase.kt
+++ b/fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/ObserveFingerprintScanStatusUseCase.kt
@@ -16,7 +16,10 @@ import kotlinx.coroutines.launch
 import javax.inject.Inject
 import javax.inject.Singleton
 import androidx.preference.PreferenceManager
+import com.simprints.infra.logging.Simber
 import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
 
 @Singleton
 class ObserveFingerprintScanStatusUseCase @Inject constructor(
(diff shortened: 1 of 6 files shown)


luhmirin-s (approved):

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/ObserveFingerprintScanStatusUseCase.kt
```diff
@@ -63,7 +69,7 @@ class ObserveFingerprintScanStatusUseCase @Inject constructor(
         observeJob = null
     }
 
-    private suspend fun turnOnFlashingWhiteSmileLeds() {
+    private suspend fun turnOnFlashingWhiteSmileLeds()  {
```
luhmirin-s, line 72: Accidental space

fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v2/channel/MainMessageChannel.kt
```diff
@@ -6,13 +6,31 @@ import com.simprints.fingerprint.infra.scanner.v2.incoming.main.MainMessageInput
...
-    incoming: MainMessageInputStream,
-    outgoing: MainMessageOutputStream
+    incoming: MainMessageInputStream, outgoing: MainMessageOutputStream
 ) : MessageChannel<MainMessageInputStream, MainMessageOutputStream>(incoming, outgoing) {
 
-    inline fun <reified R : IncomingMainMessage> sendMainModeCommandAndReceiveResponse(command: OutgoingMainMessage): Single<R> =
-        outgoing.sendMessage(command)
-            .doSimultaneously(incoming.receiveResponse())
+    // Mutex to ensure that only one thread can send and receive messages at a time
+    val mutex = Mutex()
```
luhmirin-s, line 20: Should the same apply to other channels as well?

meladRaouf: Yes, you're right; we should use the same logic for all channels. But the problem is that testing the OTA and other scanner flows will be hard, and there haven’t been any reports of concurrency issues with the other channels.

luhmirin-s: Maybe worth making a ticket to remember to look into it for upcoming major releases.

luhmirin-s (approved):

meladRaouf:
I will close this PR for now, as it is difficult to validate whether it resolves the issue. We can revisit it later when we refactor how the scanner receives commands from SID.

PR #980, opened 2024-11-19
MS-825 Ignore image transfer cancellation exception to avoid breaking user flow

luhmirin-s:
* Cancellation exception is expected when the task is cancelled, so no extra error handling is required - UI is automatically reset to a starting state.


BurningAXE:
Is this going to the release or in `main`?

luhmirin-s:
It was initially flagged as part of the release, but it is a minor issue that could go into main first.

PR #984, opened 2024-11-20
MS-802 Troubleshooting screen base

luhmirin-s:
* Implemented the base to display more complex troubleshooting information in the future. For now, it is mostly placeholders.
* The screen is hidden behind multiple clicks within a second on the "About->Sync config" in settings or the logo on the login request page. Additionally, the option is settings is protected by the settings password since it will eventually contain project-specific info. 
  * I decided against adding a feature flag check since it would not be available before login.
  * I also decided not to introduce different settings passwords - the same users who have access to the settings changes are the intended audience for the troubleshooting screens anyway. This could change in future.


meladRaouf:
Since the dashboard module is growing larger, would it be good to create a new module for this screen?

luhmirin-s:
@meladRaouf That's a good point; I will look into that. Since there are already a couple of PRs in a chain and I don't really want to rebase all of them, I will do the extraction into a module as a separate PR between phases 1 and 2 of the work.

alexandr-simprints:
Should this PR be reviewed now, or shall we wait for the module separation?

luhmirin-s:
@alexandr-simprints please review it now. Module separation should not affect much of the content of the classes, just the location.

meladRaouf (approved):

PR #985, opened 2024-11-21
MS-803 Network troubleshooting information

luhmirin-s:
* I added a section to the overview page instead of a separate page since the amount of information that we can reliably get from system services is not that large.
* Ping button requests the root of BFSID API. It is expected to return 404, but since the BFSID routing returns it, it is as good as any other endpoint to confirm that we can reach the backend.

### DIFF
--- a/infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProvider.kt
+++ b/infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProvider.kt
@@ -1,6 +1,6 @@
 package com.simprints.infra.network.url
 
-internal interface BaseUrlProvider {
+interface BaseUrlProvider {
     fun getApiBaseUrl(): String
     fun getApiBaseUrlPrefix(): String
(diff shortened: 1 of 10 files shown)


infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProvider.kt
```diff
@@ -1,6 +1,6 @@
 package com.simprints.infra.network.url
 
-internal interface BaseUrlProvider {
+interface BaseUrlProvider {
```
alexandr-simprints, line 3: Just double checking that the `internal` keyword removal is intentionaal

alexandr-simprints (approved):

infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProvider.kt
luhmirin-s: It is required in the dashboard/troubleshooting module for the ping button.

meladRaouf (approved):

PR #995, opened 2024-11-28
[MS-831] Ensure audio alerts play immediately

meladRaouf:


luhmirin-s (approved):

fingerprint/capture/src/main/java/com/simprints/fingerprint/capture/screen/ObserveFingerprintScanStatusUseCase.kt
```diff
@@ -90,23 +76,19 @@ class ObserveFingerprintScanStatusUseCase @Inject constructor(
     }
 
     private suspend fun setUiAfterScan(isGoodScan: Boolean) {
-        // Check if the previous state was ScanCompleted to avoid displaying the bad or good scan UI twice
         // There's no need to check the configuration, as the good/bad scan visual notifications apply across all LED modes.
-        if (previousState == FingerprintScanState.ScanCompleted) {
```
BurningAXE, line 95: No need to check previous state anymore?

meladRaouf: I discovered that this check was part of an old implementation where `FingerprintScanState.ScanCompleted` was being sent twice, which is not valid now.

PR #996, opened 2024-11-28
MS-804 Troubleshooting event scope log

luhmirin-s:
* Added a generic list layout and adapter that could be reused for other troubleshooting log screens.
* Created a new screen that lists all local event scopes with a "details" screen that lists all of the events in the scope.
* Added a special "toSafeString()" method to all event payload classes that expose only technical (sometimes even tokenized) data to be displayed in the log screen.

### DIFF
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/troubleshooting/adapter/TroubleshootingItemViewData.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/troubleshooting/adapter/TroubleshootingItemViewData.kt
@@ -1,0 +1,8 @@
+package com.simprints.feature.dashboard.settings.troubleshooting.adapter
+
+data class TroubleshootingItemViewData(
+    val title: String,
+    val subtitle: String = "",
+    val body: String = "",
+    val navigationId: String? = null,
+)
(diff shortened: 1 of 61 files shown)


luhmirin-s:
Would it be worth adding tests for all the "toString()" methods?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/troubleshooting/adapter/TroubleshootingItemViewData.kt
```diff
@@ -0,0 +1,8 @@
+package com.simprints.feature.dashboard.settings.troubleshooting.adapter
+
+data class TroubleshootingItemViewData(
+    val title: String,
+    val subtitle: String = "",
+    val body: String = "",
+    val navigationId: String? = null,
```
alexandr-simprints, line 7: `navigationId` is always being referenced as `scopeId` in other callers. Don't we want to rename it to `scopeId`/`eventScopeId` for consistency?

alexandr-simprints (approved):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/troubleshooting/adapter/TroubleshootingItemViewData.kt
luhmirin-s: The view data is generic on purpose so that we can reuse the same layouts and adapter combos in multiple screens.

alexandr-simprints:
> Would it be worth adding tests for all the "toSafeString()" methods?

Generally, I would advocate that it is worth it to make sure that the sensitive data isn't 'leaked' there, and to create an additional static verification for any potential changes. It doesn't fully remove the chance of sensitive data being 'leaked' but minimizes an unintentional human error

PR #998, opened 2024-11-28
[MS-565] Ensure any needed firmware is downloaded after successful login

BurningAXE:


infra/auth-logic/src/main/java/com/simprints/infra/authlogic/authenticator/ProjectAuthenticator.kt
```diff
@@ -88,4 +91,8 @@ internal class ProjectAuthenticator @Inject constructor(
             configManager.getPrivacyNotice(projectId, language).collect()
         }
     }
+
+    private suspend fun fetchFirmwareIfNeeded() {
```
luhmirin-s, line 95: A private method for a single line feels a bit redundant. 

Also, what is the "ifNeeded" part referring to? There are no checks in the method.

BurningAXE: Wanted to make it clear that it's safe to call for all project configurations. Obviously not successful :D Will replace with a comment.

luhmirin-s (approved):

PR #1003, opened 2024-12-03
Update github workflows to have release workflow

TristramN:
Main considerations:
- Version Code: I have updated the version code to be based off the run number. Remove the current version code and set a `base` and then add the run number to it. Github actions now supports supplying a sequential run number. Full breakdown in the readme. 
- Version Name: The version name now uses the github action run number and run attempt (instead of the unix time). This means we can go see the exact workflow run a build came from. It's also simpler & shorter for the delivery team and users. Full breakdown in the readme. 
- Added deployment checks. Now when a pipeline reaches a certain point (entering a new environment) it requires that deployment leads release it. This allows us to have checks before deploying the pipeline to prod (like bitbucket). 
- Added a light naming convention to the workflow files. Github doesn't allow any nesting in its file structure. So I have prefixed all reusable workflow with [Reusable] so they don't clutter the UI anymore and are all in the same place. 
- I have removed several things from gradle. The firebase deployment and the google play publish. Since they don't have anything to do with SID itself they are solely for CD. I have moved them into github actions. This speeds up their steps considerably. 
- I have unified the building steps in a reusable workflow. So we don't need to keep installing java/gradle and running builds again after the APK for each pipeline is built. 
- Updated the readme to have a workflow diagram.


.github/workflows/README.md
```diff
@@ -35,27 +68,24 @@ The CD Workflow is responsible for automatically deploying new code changes to d
...
 
 
-**Workflow Trigger and Jobs**
-
-The CD Workflow is triggered by manual trigger through workflow dispatch, allowing developers to initiate the deployment process on demand.
+**Version Code**
 
-The CD Workflow consists of several jobs, each responsible for a specific deployment task. For instance, the `deploy-to-dev` job deploys the dev build to Firebase, while the `promote-artifact` job promotes the release build to the specified Google Play track.
+The version code is generated from the sum of 2 things:
+1. The unix time in seconds / 1000. This ensures a code that is unique to that moment and always incrementing.
```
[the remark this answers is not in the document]
TristramN: You're referring to `unix time in seconds / 1000`? IE The current implementation? 

Did you read the overview? Is your preference option 2?

.github/workflows/deploy-release-workflow.yml
```diff
@@ -0,0 +1,29 @@
+name: "Deploy: Release Workflow"
+
+on:
+    workflow_dispatch:
+
+jobs:
+
```
meladRaouf, line 7: What do you think about adding a concurrency block here to prevent multiple deployments from running at the same time? Here's an example of how we currently handle it:  

```yaml
concurrency:
    group: release-internal-workflow # Only one instance of this workflow can run at a time
    cancel-in-progress: true
```

meladRaouf (approved):
Thanks for grouping the workflow files and adding reusable components! The pipeline code looks much cleaner now.
I noticed that dev, staging, and internal deployments run in parallel, which is great for efficiency. However, I’m wondering if we need to trigger all three environments every time. For example, during regular development, we mostly need builds for dev when moving tasks to UAT, and internal only at the final stages. Staging builds have also been rarely needed in the past six months.

Would it make sense to trigger deployments to these environments only when needed, to avoid unnecessary runs?

.github/workflows/deploy-release-workflow.yml
TristramN: Yes good flag! Do we want to shut down the concurrency at the root workflow itself, or when it gets to the deployment stages?
Pro it keeps it easier to follow
Con you can't set off the unit tests for multiple runs at the same time (though I'm not sure we'd want to?)

TristramN:
> Thanks for grouping the workflow files and adding reusable components! The pipeline code looks much cleaner now. I noticed that dev, staging, and internal deployments run in parallel, which is great for efficiency. However, I’m wondering if we need to trigger all three environments every time. For example, during regular development, we mostly need builds for dev when moving tasks to UAT, and internal only at the final stages. Staging builds have also been rarely needed in the past six months.
> 
> Would it make sense to trigger deployments to these environments only when needed, to avoid unnecessary runs?

@meladRaouf good question! Firstly I don't think we'd use this workflow for normal development right? That's what the PR workflow is for. This one would only run on `release/*` branches when manually deployed. Unless I am misunderstanding something? 

For running the workflow on releases, in terms of them running in parallel are we thinking:
1. Do it the similar way to when we had bitbucket pipelines and have them all run sequentially when we release the next stage? Specifically it ran tests and then build debug, and then when someone would trigger the next step and it would build and deploy staging, etc. 
2. Keep the pipelines parallel to each other, but only proceed with them when someone approves that step?

.github/workflows/deploy-release-workflow.yml
TristramN: Having thought about this more I'm pro shutting it down at the root. This will be way cleaner for making sure the latest run has the latest version code.

.github/workflows/README.md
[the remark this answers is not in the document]
luhmirin-s: Are we already doing that? :O 

Do we need to truncate it in the first place? maybe we could at least do some more meaningful division - 3600 (per hour) or 900 (per 15 minutes)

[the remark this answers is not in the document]
TristramN: Yep the current version code is derived from `unix time in seconds / 1000`. For example the currently deployed version code is `1732785`. 

It used to be the bitbucket pipeline number. My understanding is that when we moved to github, github actions at the time did not have the equivalent ability to have a sequentially increasing run number, so it was decided to use a unix timestamp. Which while it works, isn't super useful. 

Now thankfully github actions does have the same feature. Hence my proposal in option 2, because at least this way the version code would mean something, we could take it and point to the specific run that built the deployed release. 

And yes, if we kept option 1 we could truncate it differently! Which it will note down. Though it would always be limited in usefulness. Ex: we can't take `1732785` x 1000 and know exactly when it was built haha.

TristramN:
@meladRaouf Option 2 is implemented here: https://github.com/Simprints/Android-Simprints-ID/blob/devops/option-2-version-code/.github/workflows/README.md?plain=1#L71-L78

IMO this is also my preference. Let me know if you have any thoughts otherwise I will merge into this PR.

BurningAXE:
> Option 2) Take the current version code (xxxxxxx) and set it as the base and then add the run number to it. We could also jump to a simpler base initially, like 9000000 which is higher than the current version, and then add the run number. So 9000054 would mean workflow https://github.com/Simprints/Android-Simprints-ID/pull/54 deployed this version.

I, too, prefer Option 2. However, do we need to jump straight to 9000000 instead of 2000000? If we go with 9000000 and some other migration/refactoring in the future forces us to start from a new base in the future, we'd need to add a new digit.

BurningAXE:
A side note - you'll probably do it without my prompt but please please squash related commits before merge :D

.github/workflows/README.md
```diff
@@ -35,27 +68,27 @@ The CD Workflow is responsible for automatically deploying new code changes to d
...
-The CD Workflow is triggered by manual trigger through workflow dispatch, allowing developers to initiate the deployment process on demand.
-
-The CD Workflow consists of several jobs, each responsible for a specific deployment task. For instance, the `deploy-to-dev` job deploys the dev build to Firebase, while the `promote-artifact` job promotes the release build to the specified Google Play track.
-
-## **Dependency Updates workflow**
+**Version Code**
 
-Updates project dependencies using Dependabot, an automated dependency management tool, ensuring that the project always uses the latest stable versions of its dependencies.
+The version code is generated from 3 things:
+1. A "base" version code of `10000000`. For a time the version code was derived from unix time / 1000. This created a always increasing number, but was otherwise not so useful. Setting a base allows us to establish a clean "floor" for the version code. If the workflow runs ever exceed the base, first of all go us, and 2nd the base can be removed.
```
BurningAXE, line 74: Ah, you went with 10000000 instead of 9000000. That's longer but kind of "cleaner".

.github/workflows/README.md
```diff
@@ -35,27 +68,27 @@ The CD Workflow is responsible for automatically deploying new code changes to d
...
-
-## **Dependency Updates workflow**
+**Version Code**
 
-Updates project dependencies using Dependabot, an automated dependency management tool, ensuring that the project always uses the latest stable versions of its dependencies.
+The version code is generated from 3 things:
+1. A "base" version code of `10000000`. For a time the version code was derived from unix time / 1000. This created a always increasing number, but was otherwise not so useful. Setting a base allows us to establish a clean "floor" for the version code. If the workflow runs ever exceed the base, first of all go us, and 2nd the base can be removed. 
+2. The `run number` of the workflow. This tells us which run a build came from and is auto incrementing. 
+3. The last two digits are reserved for the `run attempts`. This tells us how many times a specific workflow was run. If more than 99 attempts are made (something is wrong) the build will fail and you need to start a new workflow run.
```
BurningAXE, line 76: Do we really need this? If a workflow fails, we'll either do 1-2 more attempts to make sure it was not a fluctuation or fix an issue and do another workflow run. In any case the build will always come from the last workflow attempt. Or am I missing something?

.github/workflows/README.md
```diff
@@ -35,27 +68,27 @@ The CD Workflow is responsible for automatically deploying new code changes to d
...
 
-To promote code reusability and efficiency, two reusable workflows are defined:
+**Version Name**
 
-1.  **Test Android Modules:** This workflow takes a list of modules and a unique report ID as input and executes unit tests for the specified modules. It then uploads the test coverage reports using the provided report ID.
+The version name follows our versioning convention:
+- `year`.`quarter`.`release`-`(optional) deployment`+`(optional) unix timestamp`.`run number`.`run attempt`
 
-2.  **Deploy to Firebase:** This workflow takes a build type (dev or staging) as input and uploads the corresponding APK (debug or staging) to the Firebase distribution track.
+The optional params are **only** used on none release builds. We add the unix timestamp so we can compute the `versionCode` from the filename if needed.
```
BurningAXE, line 85: That's no longer valid, right? 
- `year`.`quarter`.`release`-`(optional) deployment`+`(10000000+run number+run attempt)`

.github/workflows/reusable-promote-artifact.yml
```diff
@@ -0,0 +1,80 @@
...
+
+            -   name: Set vars
+                run: |
+                    if [ "${{inputs.deployment-track}}" == "Internal" ]; then
+                        STATUS="completed"
+                        USER_FRACTION=1.0
+                        TRACK="qa"
+                    elif [ "${{inputs.deployment-track}}" == "Alpha" ]; then
+                        STATUS="completed"
+                        USER_FRACTION=0.33
```
BurningAXE, line 55: Always deploy to 33% of Alpha? :)

.github/workflows/pipeline-deploy-to-internal.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: ./.github/workflows/reusable-promote-artifact.yml
+        with:
+            deployment-track: internal
+            upload-artifact: ${{ needs.build-internal-aab.outputs.build-artifact }}
+            mapping-file: ${{ needs.build-internal-aab.outputs.optional-mapping-file }}
+        needs:
+            - build-internal-aab
+            - run-app-sweep
+
+    deploy-alpha-build:
```
BurningAXE, line 33: Hm, do we always want to deploy simultaneously to both Internal and Alpha? Looking at the lists of both - I'm not sure what's the distinction (and hence use of Alpha) currently.

.github/workflows/reusable-build-apk.yml
```diff
@@ -0,0 +1,139 @@
...
+
+            -   name: Check base version code
+                if: ${{ github.run_number * 100 >= env.BASE_VERSION_CODE }}
+                run: |
+                    echo "Github workflows now exceeds the base version code. Please remove the BASE_VERSION_CODE"
+                    echo "BASE_VERSION_CODE=0" >> $GITHUB_ENV
+
+            -   name: Compute version variables
+                run: |
+                    echo "VERSION_CODE=$(($BASE_VERSION_CODE + ${{ github.run_number }} * 100 + ${{ github.run_attempt }}))" >> $GITHUB_ENV
```
BurningAXE, line 66: Hm, I'm confused - is it going to be `run_number*100+run_attempt` or `run_number`.`run_attempt`?
Ie. if run_number is 123 and run_attempt is 45:
`12345` vs `123.45`

.github/workflows/README.md
TristramN: It's a good questions @BurningAXE. I did it because it would prevent deployment collisions and make it explicit where exactly the build came from. Without it if someone runs the pipeline and it succeeds then for some reason the pipeline is re-run again (which is doable in the github UI) the re-run would fail the deployment. And if someone looked at the build name they'd come to the pipeline run but then have to search through to find which attempt it came from. 

 Those are small things though. So perhaps is just best to drop it for simplicity?

.github/workflows/reusable-promote-artifact.yml
TristramN: Haha good catch. I don't even think Alpha has the rollout ability? I will remove this.

.github/workflows/pipeline-deploy-to-internal.yml
TristramN: This won't deploy them simultaneously. Since Alpha is a protected environment what it will do is trigger the request to deploy to alpha. But the step isn't run until it's approved.

.github/workflows/reusable-build-apk.yml
TristramN: The version code has to be an int. So the version name would be `123.45` but the version code will be `12345`

.github/workflows/README.md
TristramN: The unix date is no longer valid good point! But also we don't currently put the "base" number in the version name, only the version code. 

Version name = 2024.1.0+15.2
Version code = 100001502

Since the version name is user and delivery facing I thought `2024.1.0+15.2` was more friendly than `2024.1.0+100001502`. You can still compute the version code from the version name if we need to.

luhmirin-s (approved):

PR #1046, opened 2025-01-06
MS-846 Safe navigation fixes

luhmirin-s:
* Fixed the incorrect "current" fragment in sync/main fragment. Also checked the rest of `navigateSafely()` calls to ensure that the provided fragment matches the one used in the graph.
* Fixed a related issue in the troubleshooting fragment in a bit more complex way. Fragments created as a part of the pager adapter do not have proper access to the "parent/root" fragment, so the navigation is done via the shared view model instead.
* Switched to navigation using "directions" where possible to add a layer of type safety for the basic app navigation.

As a bonus, I have found and fixed a minor issue with the project config not scheduling a one-time worker due to using the name of the periodic one.

### DIFF
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/main/sync/SyncFragment.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/main/sync/SyncFragment.kt
@@ -10,6 +10,7 @@ import androidx.navigation.fragment.findNavController
 import com.simprints.core.livedata.LiveDataEventWithContentObserver
 import com.simprints.feature.dashboard.R
 import com.simprints.feature.dashboard.databinding.FragmentDashboardCardSyncBinding
+import com.simprints.feature.dashboard.main.MainFragmentDirections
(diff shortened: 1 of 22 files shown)


feature/dashboard/src/main/java/com/simprints/feature/dashboard/main/sync/SyncFragment.kt
alexandr-simprints, line 1: Is there are reason why `R.id.action_mainFragment_to_requestLoginFragment` and `R.id.action_mainFragment_to_login` invocations were not changed to `MainFragmentDirections.*`?

alexandr-simprints (approved):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/main/sync/SyncFragment.kt
luhmirin-s: Bundle args don't work with navigation by direction without "workarounds".

PR #1047, opened 2025-01-07
Adding progress indicator for image transfer

alexandr-simprints:
Adding circular loading indicator that becomes visible during the image transfer

https://github.com/user-attachments/assets/42c8caa4-842f-4208-809e-92630f893ae6

https://github.com/user-attachments/assets/d3cfe4b8-b979-4d3f-bb02-afc8767a6f3d


luhmirin-s (approved):

BurningAXE:
Hm, that's a little weird UX IMO. I preferred the old one even if the progress bar seemed frozen at the end. Maybe a better solution would be to increase the time it takes to reach the end as image transfer takes ± constant time for each resolution.

PR #1055, opened 2025-01-13
MS-676 Sync config button upgrades

luhmirin-s:
As per updated requirements:
* Added refresh icon (see video below).
* Added a toast notification after both sync workers finish. Since "start" and "end" toasts were shown way too quickly, I decided to change the description text under the preference instead of the "starting" toast (see video below).
* I added a subtitle with the time since the "last update." It is in relative format to keep it consistent with the "last sync" indicator.

https://github.com/user-attachments/assets/48f0b347-b75b-4eb9-9d1a-544d621f4b7e

* As a bonus, I have also added some details about the current configuration to the troubleshooting overview - config ID, config update data (in vulcan) and time since last config sync.

### DIFF
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
@@ -109,10 +109,6 @@ internal class DebugFragment : Fragment(R.layout.fragment_debug) {
             }
         }
 
-        binding.syncDevice.setOnClickListener {
-            syncOrchestrator.startDeviceSync()
-        }
-
         binding.printRoomDb.setOnClickListener {
             binding.logs.text = ""
             runBlocking {
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
@@ -14,7 +14,6 @@ import com.simprints.core.DispatcherIO
 import com.simprints.feature.dashboard.R
 import com.simprints.feature.dashboard.databinding.FragmentDebugBinding
 import com.simprints.infra.authstore.AuthStore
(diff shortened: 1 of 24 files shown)


jonsimprints:
> As per updated requirements:
> 
> * Added refresh icon (see video below).
> * Added a toast notification after both sync workers finish. Since "start" and "end" toasts were shown way too quickly, I decided to change the description text under the preference instead of the "starting" toast (see video below).
> * I added a subtitle with the time since the "last update." It is in relative format to keep it consistent with the "last sync" indicator.
> 
>  Screen_recording_20250113_175317.mp4 
> * As a bonus, I have also added some details about the current configuration to the troubleshooting overview - config ID, config update data (in vulcan) and time since last config sync.
> 
> 

This looks brilliabt

feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
```diff
@@ -109,10 +109,6 @@ internal class DebugFragment : Fragment(R.layout.fragment_debug) {
             }
         }
```
meladRaouf, line 111: Do we still need the config button here?

meladRaouf (approved):

luhmirin-s:
1. Removed password requirement
2. Removed the random parenthesis from the network

I am unsure that animation is worth the hassle, considering that it would run for a few seconds and is within the performance fragment (I had to implement a custom layout just to add an icon on the right 😄).

feature/dashboard/src/main/java/com/simprints/feature/dashboard/debug/DebugFragment.kt
luhmirin-s: Not really. Removed.

luhmirin-s:
1. Removed password requirement
2. Removed the random parenthesis from the network

I am unsure that animation is worth the hassle, considering that it would run for a few seconds and is within the performance fragment (I had to implement a custom layout just to add an icon on the right 😄).

PR #1056, opened 2025-01-13
[MS-821] Update Enrol+ to capture all but match with Matching Modalities only

BurningAXE:


feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCaseTest.kt
```diff
@@ -257,6 +318,27 @@ class BuildStepsUseCaseTest {
...
+        every { action.getSubjectAgeIfAvailable() } returns null
+        every { cache.steps } returns listOf(
+            Step(StepId.FACE_CAPTURE, mockk(relaxed = true), mockk(relaxed = true), mockk(relaxed = true)),
+        )
+
+        val steps = useCase.build(action, projectConfiguration)
+
+        assertStepOrder(
+            steps,
+            StepId.FINGERPRINT_CAPTURE,
```
meladRaouf, line 336: Why are there two fingerprint capture steps here? I don't understand.

meladRaouf (approved):

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCaseTest.kt
luhmirin-s: I was also confused and digging - it seems the "common configuration" includes 2 SDKs for fingerprint, therefore a step per SDK.

luhmirin-s (approved):

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCaseTest.kt
BurningAXE: Exactly! The test configuration has two fingerprint SDKs with overlapping ages - something that at least the step builder supports so it's being tested as well.

PR #1061, opened 2025-01-16
[MS-716] Face auto-capture

alex-vt:
# Face auto-capture

Jira: MS-716

## UX

Same as the regular manual capture, except:

#### 1. When face image of good quality appears in viewfinder:

| for manual capture: | for auto-capture: |
|---|---|
| Green button appears - press to capture | Capture starts automatically and completes in 2 seconds |

#### 2. Initial hold-off, to let user aim the viewfinder:

* First 2 seconds when viewfinder starts
* Also if user clicks away from Simprints ID and returns

## Configuration

> `<Project>` -> `Configurations` -> `Custom`

Face auto-capture configuration example - enabled for device1 and device2:

```json
"faceAutoCapture": {
    "enabled": true,
    "deviceIdFilter": ["device1", "device2"]
}
```

If no `deviceIdFilter`: entire device pool eligible.

If no `faceAutoCapture` config at all: no auto-capture.

## Event data

Same as for regular manual capture, except:
* `isAutoCapture: true` included with the capture events
* Events registered only for qualifying images

#### Live feedback view & viewmodel

The `livefeedbackautocapture` package is similar to the conventional capture `livefeedback` one, except:
* Auto-capture has a preparation period from the moment `LiveFeedbackAutoCaptureFragment` becomes visible and for `LiveFeedbackAutoCaptureFragmentViewModel.AUTO_CAPTURE_VIEWFINDER_RESUME_DELAY_MS` amount of time. This helps prevent startling the user with a too early start of auto-capture - before the user would aim the viewfinder at a face.
* The auto-capture imaging progress starts when a qualifying face image appears in the viewfinder, and lasts for `LiveFeedbackAutoCaptureFragmentViewModel.AUTO_CAPTURE_IMAGING_DURATION_MS` amount of time.
* The number of images auto-captured is limited to `LiveFeedbackAutoCaptureFragmentViewModel.AUTO_CAPTURE_SAMPLE_COUNT`. Imaging is done by sampling the frames over even periods of time within the auto-capture imaging time period.
* The number of images to accept is `LiveFeedbackAutoCaptureFragmentViewModel.samplesToKeep` - similarly to `LiveFeedbackFragmentViewModel.samplesToCapture`. The best quality sampled images are selected to keep.
* The kept sampled images and the fallback image are included in the capture events - unlike in `LiveFeedbackFragmentViewModel` where all conventionally captured and the fallback images are included.
* The capture events have an `isAutoCapture` flag value of `true`.

The reason for having this package as a near-copy of `livefeedback` is to keep both implementations independent but similar. The layout for `LiveFeedbackAutoCaptureFragment` is `fragment_live_feedback_auto_capture.xml`.

#### Decision-making between manual and auto-capture

If the project is configured to have auto-capture enabled on the current device ID (or on all devices):
* An `isAutoCapture` flag, included into `face/capture/src/main/java/com/simprints/face/capture/FaceCaptureContract` at the face capture step at the orchestrator in `feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCase`, in enabled.
* The `face/capture/src/main/java/com/simprints/face/capture/screens/controller/FaceCaptureControllerFragment` reads the auto-capture flag from the `FaceCaptureContract` and selects the appropriate navigation graph.
* The auto-capture `graph_face_capture_auto_internal` navigation graph is similar to the manual `graph_face_capture_internal` one, except the regular live feedback fragment is replaced with the similar auto-capture one (see description above).

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragment.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragment.kt
@@ -1,0 +1,367 @@
+package com.simprints.face.capture.screens.livefeedbackautocapture
+
+import android.Manifest
+import android.content.Intent
+import android.graphics.Bitmap
+import android.net.Uri
+import android.os.Bundle
+import android.provider.Settings
+import android.util.Size
+import android.view.View
+import androidx.activity.result.contract.ActivityResultContracts
+import androidx.camera.core.CameraSelector.DEFAULT_BACK_CAMERA
+import androidx.camera.core.ImageAnalysis
+import androidx.camera.core.ImageAnalysis.OUTPUT_IMAGE_FORMAT_RGBA_8888
+import androidx.camera.core.Preview
+import androidx.camera.lifecycle.ProcessCameraProvider
+import androidx.camera.lifecycle.awaitInstance
+import androidx.core.content.ContextCompat
+import androidx.core.view.isGone
+import androidx.core.view.isVisible
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.activityViewModels
+import androidx.fragment.app.viewModels
+import androidx.lifecycle.lifecycleScope
+import androidx.navigation.fragment.findNavController
+import com.simprints.core.domain.permission.PermissionStatus
+import com.simprints.core.tools.extentions.hasPermission
+import com.simprints.core.tools.extentions.permissionFromResult
+import com.simprints.face.capture.R
+import com.simprints.face.capture.databinding.FragmentLiveFeedbackBinding
+import com.simprints.face.capture.models.FaceDetection
+import com.simprints.face.capture.screens.FaceCaptureViewModel
+import com.simprints.face.capture.screens.livefeedback.CropToTargetOverlayAnalyzer
+import com.simprints.infra.logging.Simber
+import com.simprints.infra.uibase.navigation.navigateSafely
+import com.simprints.infra.uibase.view.setCheckedWithLeftDrawable
+import com.simprints.infra.uibase.viewbinding.viewBinding
+import dagger.hilt.android.AndroidEntryPoint
+import kotlinx.coroutines.launch
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import com.simprints.infra.resources.R as IDR
+
+/**
+ * As the user is capturing subject's face, they are presented with this fragment, which displays
+ * live information about distance and whether the face is ready to be captured or not.
+ * It also displays the capture process of the face and then sends this result to
+ * [com.simprints.face.capture.screens.confirmation.ConfirmationFragment]
+ */
+@AndroidEntryPoint
+internal class LiveFeedbackAutoCaptureFragment : Fragment(R.layout.fragment_live_feedback_auto_capture) {
+    /** Blocking camera operations are performed using this executor */
+    private lateinit var cameraExecutor: ExecutorService
+
+    private val mainVm: FaceCaptureViewModel by activityViewModels()
+
+    private val vm: LiveFeedbackAutoCaptureFragmentViewModel by viewModels()
+    private val binding by viewBinding(FragmentLiveFeedbackBinding::bind)
+
+    private lateinit var screenSize: Size
+    private lateinit var targetResolution: Size
+
+    private val launchPermissionRequest = registerForActivityResult(
+        ActivityResultContracts.RequestPermission(),
+    ) { granted ->
+        when (requireActivity().permissionFromResult(Manifest.permission.CAMERA, granted)) {
+            PermissionStatus.Granted -> setUpCamera()
+            PermissionStatus.Denied -> renderNoPermission(false)
+            PermissionStatus.DeniedNeverAskAgain -> renderNoPermission(true)
+        }
+    }
+
+    override fun onViewCreated(
+        view: View,
+        savedInstanceState: Bundle?,
+    ) {
+        super.onViewCreated(view, savedInstanceState)
+        initFragment()
+    }
+
+    private fun initFragment() {
+        screenSize = with(resources.displayMetrics) { Size(widthPixels, widthPixels) }
+        bindViewModel()
+        binding.captureProgress.max = 1 // normalized progress
+
+        // Wait till the views gets its final size then init frame processor and setup the camera
+        binding.faceCaptureCamera.post {
+            if (view != null) {
+                vm.initCapture(mainVm.samplesToCapture, mainVm.attemptNumber)
+            }
+        }
+    }
+
+    /** Initialize CameraX, and prepare to bind the camera use cases  */
+    private fun setUpCamera() = lifecycleScope.launch {
+        if (::cameraExecutor.isInitialized && !cameraExecutor.isShutdown) {
+            return@launch
+        }
+        vm.startPreparationDelay()
+
+        // Initialize our background executor
+        cameraExecutor = Executors.newSingleThreadExecutor()
+        // ImageAnalysis
+        // Todo choose accurate output image resolution that respects quality,performance and face analysis SDKs https://simprints.atlassian.net/browse/CORE-2569
+        if (!::targetResolution.isInitialized) {
+            targetResolution = Size(binding.captureOverlay.width, binding.captureOverlay.height)
+        }
+
+        val imageAnalyzer = ImageAnalysis
+            .Builder()
+            .setTargetResolution(targetResolution)
+            .setOutputImageRotationEnabled(true)
+            .setOutputImageFormat(OUTPUT_IMAGE_FORMAT_RGBA_8888)
+            .build()
+        val cropAnalyzer = CropToTargetOverlayAnalyzer(binding.captureOverlay, ::analyze)
+
+        imageAnalyzer.setAnalyzer(cameraExecutor, cropAnalyzer)
+
+        // Preview
+        val preview = Preview.Builder().setTargetResolution(targetResolution).build()
+        val cameraProvider = ProcessCameraProvider.awaitInstance(requireContext())
+        cameraProvider.unbindAll()
+        cameraProvider.bindToLifecycle(
+            this@LiveFeedbackAutoCaptureFragment,
+            DEFAULT_BACK_CAMERA,
+            preview,
+            imageAnalyzer,
+        )
+        // Attach the view's surface provider to preview use case
+        preview.surfaceProvider = binding.faceCaptureCamera.surfaceProvider
+    }
+
+    override fun onResume() {
+        super.onResume()
+
+        when {
+            requireActivity().hasPermission(Manifest.permission.CAMERA) -> setUpCamera()
+            mainVm.shouldCheckCameraPermissions.getAndSet(false) -> {
+                // Check permission in onResume() so that if user left the app to go to Settings
+                // and give the permission, it's reflected when they come back to SID
+                if (requireActivity().hasPermission(Manifest.permission.CAMERA)) {
+                    setUpCamera()
+                } else {
+                    launchPermissionRequest.launch(Manifest.permission.CAMERA)
+                }
+            }
+
+            else -> mainVm.shouldCheckCameraPermissions.set(true)
+        }
+    }
+
+    override fun onStop() {
+        // Shut down our background executor
+        if (::cameraExecutor.isInitialized) {
+            cameraExecutor.shutdown()
+        }
+        super.onStop()
+    }
+
+    private fun bindViewModel() {
+        vm.currentDetection.observe(viewLifecycleOwner) {
+            renderCurrentDetection(it)
+        }
+
+        vm.capturingState.observe(viewLifecycleOwner) {
+            @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA") when (it) {
+                LiveFeedbackAutoCaptureFragmentViewModel.CapturingState.NOT_STARTED -> renderCapturingNotStarted()
+
+                LiveFeedbackAutoCaptureFragmentViewModel.CapturingState.CAPTURING -> renderCapturing()
+
+                LiveFeedbackAutoCaptureFragmentViewModel.CapturingState.FINISHED -> {
+                    mainVm.captureFinished(vm.sortedQualifyingCaptures)
+                    findNavController().navigateSafely(
+                        currentFragment = this,
+                        directions = LiveFeedbackAutoCaptureFragmentDirections.actionFaceLiveFeedbackFragmentToFaceConfirmationFragment(),
+                    )
+                }
+            }
+        }
+    }
+
+    private fun analyze(image: Bitmap) {
+        try {
+            vm.process(croppedBitmap = image)
+        } catch (t: Throwable) {
+            Simber.e("Image analysis crashed", t)
+            // Image analysis is running in bg thread
+            lifecycleScope.launch {
+                mainVm.submitError(t)
+            }
+        }
+    }
+
+    private fun renderCurrentDetection(faceDetection: FaceDetection) {
+        when (faceDetection.status) {
+            FaceDetection.Status.NOFACE -> renderNoFace()
+            FaceDetection.Status.OFFYAW -> renderFaceNotStraight()
+            FaceDetection.Status.OFFROLL -> renderFaceNotStraight()
+            FaceDetection.Status.TOOCLOSE -> renderFaceTooClose()
+            FaceDetection.Status.TOOFAR -> renderFaceTooFar()
+            FaceDetection.Status.BAD_QUALITY -> renderBadQuality()
+            FaceDetection.Status.VALID -> renderValidFace()
+            FaceDetection.Status.VALID_CAPTURING -> renderValidCapturingFace()
+        }
+    }
+
+    private fun renderCapturingStateColors() {
+        with(binding) {
+            captureOverlay.drawWhiteTarget()
+            captureFeedbackTxtExplanation.setTextColor(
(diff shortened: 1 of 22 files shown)


alex-vt:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=Simprints_Android-Simprints-ID&pullRequest=1061) **Quality Gate failed**
> Failed conditions  [8.5% Duplication on New Code](https://sonarcloud.io/component_measures?id=Simprints_Android-Simprints-ID&pullRequest=1061&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=Simprints_Android-Simprints-ID&pullRequest=1061)

Live feedback package duplicated for auto-capture by design - see description above.

BurningAXE:
Nice! Appreciate this is still in draft so just some high level comments:

1. It might be worth exposing the configuration to the users so they can decide whether to use the auto-capture based on their preferences and needs. We can cascade it so that the project config determines whether the feature is available for the device at all and then the user can decide to turn it off/on. I guess this is more a question for @jonsimprints that we can take to Slack
2. Given that we are still in experimental phase and there might be several versions of auto-capture that we A/B test, it will be good to include the auto-capture version in the event. This will help us when analyzing how each one performs.

> Auto-capture has a preparation period from the moment LiveFeedbackAutoCaptureFragment becomes visible and for LiveFeedbackAutoCaptureFragmentViewModel.AUTO_CAPTURE_VIEWFINDER_RESUME_DELAY_MS amount of time. This helps prevent startling the user with a too early start of auto-capture - before the user would aim the viewfinder at a face.
This is a personal opinion and I am biased but IMHO it's better to let the user indicate when they are ready instead of triggering automatically. Even with a delay often times they might not be ready pointing the camera at the subject - say the child is looking (or even running!) away and the user needs to attract their attention.

alex-vt:
> 1. It might be worth exposing the configuration to the users so they can decide whether to use the auto-capture based on their preferences and needs. We can cascade it so that the project config determines whether the feature is available for the device at all and then the user can decide to turn it off/on. I guess this is more a question for @jonsimprints that we can take to Slack

Sure, I've left a message there and we can discuss. There are some ways to make the toggle but we should be careful to not create additional burden. The current approach is, to experiment with the feature on an initial, limited pool of specific devices - so those users would be in a good feedback loop anyway.

> 2. Given that we are still in experimental phase and there might be several versions of auto-capture that we A/B test, it will be good to include the auto-capture version in the event. This will help us when analyzing how each one performs.

Good idea thanks!

> IMHO it's better to let the user indicate when they are ready instead of triggering automatically. Even with a delay often times they might not be ready pointing the camera at the subject - say the child is looking (or even running!) away and the user needs to attract their attention.

After using this for a while I can tell that removal of the button press feels like a more smooth UX. After disabling the auto-capture that button click becomes noticeable. But sure there is a balance of convenience and control. We can discuss this as well.

BurningAXE:
> Face auto-capture configuration example - enabled for device1 and device2:
Vulcan has the option to serve different configurations to different devices:
https://simprints.atlassian.net/67fd80fd-5a2a-4afa-b7a1-6e7f72f5d22f#media-blob-url=true&id=7071d41d-00d0-48cc-b22d-4c9c97b8afef&contextId=1331888208&collection=contentId-1331888208
Do we need a separate filter in the custom config?

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
```diff
@@ -0,0 +1,261 @@
...
+    private fun isEligibleForImagingSample(currentTimestamp: Timestamp): Boolean {
+        if (userCaptures.isEmpty()) {
+            return true
+        }
+        if (userCaptures.size >= AUTO_CAPTURE_SAMPLE_COUNT) {
+            return false
+        }
+        val samplingStepMillis = AUTO_CAPTURE_IMAGING_DURATION_MS.toDouble() / AUTO_CAPTURE_SAMPLE_COUNT
+        val minNextSampleTimeMillis = userCaptures.first().detectionStartTime.ms + samplingStepMillis * userCaptures.size
+        return currentTimestamp.ms >= minNextSampleTimeMillis
```
BurningAXE, line 136: With this approach we might be missing out on good captures just because they arrived at the wrong time.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/FaceAutoCaptureEligibilityUseCase.kt
```diff
@@ -0,0 +1,44 @@
...
+internal class FaceAutoCaptureEligibilityUseCase @Inject constructor(
+    @DeviceID private val deviceID: String,
+) {
+
+    /*
+        Face auto-capture configuration example - enabled for device1 and device2:
+        ```json
+        "faceAutoCapture": {
+            "enabled": true,
+            "deviceIdFilter": ["device1", "device2"]
```
luhmirin-s, line 15: This is redundant since the custom configuration can be applied by-device already. I would prefer the configurations to be as simple as possible.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragment.kt
```diff
@@ -0,0 +1,367 @@
...
+import com.simprints.infra.resources.R as IDR
+
+/**
+ * As the user is capturing subject's face, they are presented with this fragment, which displays
+ * live information about distance and whether the face is ready to be captured or not.
+ * It also displays the capture process of the face and then sends this result to
+ * [com.simprints.face.capture.screens.confirmation.ConfirmationFragment]
+ */
+@AndroidEntryPoint
+internal class LiveFeedbackAutoCaptureFragment : Fragment(R.layout.fragment_live_feedback_auto_capture) {
```
luhmirin-s, line 51: Fine for the experiment, but eventually it would be great to abstract the capture strategy to reduce UI code duplication. All of the camera-related configuration is finicky enough, would we nice to avoid having to support 2 copies of it.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
```diff
@@ -0,0 +1,261 @@
...
+     */
+    fun process(croppedBitmap: Bitmap) {
+        val captureStartTime = timeHelper.now()
+        val potentialFace = faceDetector.analyze(croppedBitmap)
+
+        val faceDetection = getFaceDetectionFromPotentialFace(croppedBitmap, potentialFace)
+        faceDetection.detectionStartTime = captureStartTime
+        faceDetection.detectionEndTime = timeHelper.now()
+
+        if (autoCaptureInitialDelayJob.isCompleted) {
```
BurningAXE, line 83: If delay is not complete, the `faceDetection` is unused. Better to put the if in the beginning of the function.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
```diff
@@ -0,0 +1,261 @@
...
+
+    /**
+     * Since events are saved in a blocking way in [SimpleCaptureEventReporter.addCaptureEvents],
+     * to speed things up this method creates multiple async jobs and run them all in parallel.
+     *
+     * Auto-capture generates a number of sample captures that typically significantly exceeds samplesToKeep.
+     * Thus the non-qualifying samples are excluded to avoid excessive event data.
+     */
+    private fun sendQualifyingAndFallbackCaptureEvents(attemptNumber: Int) = runBlocking {
+        userCaptures
```
BurningAXE, line 232: Can't we use `sortedQualifyingCaptures` here instead of repeating the same operations on `userCaptures`?

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragment.kt
alex-vt: Makes sense - added to the backlog.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
```diff
@@ -0,0 +1,261 @@
...
+    )
+    private val fallbackCaptureEventStartTime = timeHelper.now()
+    private var shouldSendFallbackCaptureEvent: AtomicBoolean = AtomicBoolean(true)
+    private var fallbackCapture: FaceDetection? = null
+
+    val userCaptures = mutableListOf<FaceDetection>()
+    var sortedQualifyingCaptures = listOf<FaceDetection>()
+    val currentDetection = MutableLiveData<FaceDetection>()
+    val capturingState = MutableLiveData(CapturingState.NOT_STARTED)
+    private var autoCaptureInitialDelayJob: Job = viewModelScope.launch {
```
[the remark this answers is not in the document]
alex-vt: It yields to the rest of the initialization before starting, so, no delay.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
alex-vt: The logic of qualified captures and total events is slightly different, but anyway, after the frame pacing part update, it's simplified to how it is for the manual capture.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
alex-vt: Originally the potentially large volume of captured data (incl. bitmaps) and sending all of it in events was a problem big enough enough to seek a sparse solution. But as the events are sent only for the qualifying captures anyway, indeed this could be optimized, thanks! Now every frame is considered - in a positive decision case it's added if we don't have the needed quantity yet, or replaces the worst one if we do. Sorting is avoided until the capture time is over.

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/FaceAutoCaptureEligibilityUseCase.kt
alex-vt: Thanks for the suggestion - simplified to just true/false.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedbackautocapture/LiveFeedbackAutoCaptureFragmentViewModel.kt
```diff
@@ -0,0 +1,267 @@
...
+
+        if (autoCaptureInitialDelayJob.isCompleted) {
+            currentDetection.postValue(faceDetection)
+            if (faceDetection.status == FaceDetection.Status.VALID
+                && capturingState.value == CapturingState.NOT_STARTED) {
+                capturingState.postValue(CapturingState.CAPTURING)
+                captureImagingStartTime = captureStartTime.ms
+                autoCaptureImagingTimeoutJob = viewModelScope.launch {
+                    delay(AUTO_CAPTURE_IMAGING_DURATION_MS)
+                    finishCapture(attemptNumber)
```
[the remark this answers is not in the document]
alex-vt: Another idea is, auto-rejecting an incomplete sample set. Showing the user something like "The person moved, please try again" and returning to the viewfinder. It still creates some momentary UX friction but avoids getting the user stuck.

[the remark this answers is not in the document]
BurningAXE: IMO if the project is configured to capture an "impossible" number of samples, this should be addressed by updating the configuration to a more realistic number, not by designing the algorithm to always be satisfied with anything it can get. Imagine a WLDS scenario where we do want 10 samples from each subject.
As for the progress UX - it's up for debate but I'd make it depend on the determining factor. Currently that's time so it reflects it correctly. However, if we change the determining factor to be number of samples to keep, I'd put that in the progress bar.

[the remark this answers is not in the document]
BurningAXE: > Another idea is, auto-rejecting an incomplete sample set. Showing the user something like "The person moved, please try again" and returning to the viewfinder. It still creates some momentary UX friction but avoids getting the user stuck.

This will result in losing the samples captured so far and ultimately making the capture process harder (similar to what we currently have for fingerprint).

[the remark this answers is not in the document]
alex-vt: Added MS-858 for the sample completeness issue.

feature/orchestrator/src/test/java/com/simprints/feature/orchestrator/usecases/steps/BuildStepsUseCaseTest.kt
```diff
@@ -315,27 +257,6 @@ class BuildStepsUseCaseTest {
...
-        val steps = useCase.build(action, projectConfiguration)
-
-        assertStepOrder(
-            steps,
-            StepId.FINGERPRINT_CAPTURE,
-            StepId.FINGERPRINT_CAPTURE,
-            StepId.ENROL_LAST_BIOMETRIC,
-        )
-    }
-
```
BurningAXE, line 341: Why were these tests deleted?

alex-vt: Accedentally after BuildStepsUseCase changes were reverted due to the change in auto-capture config handling. Thanks for catching! Restored.

infra/config-store/src/main/java/com/simprints/infra/config/store/models/ExperimentalProjectConfiguration.kt
```diff
@@ -25,9 +25,22 @@ data class ExperimentalProjectConfiguration(
...
+
     companion object {
         internal const val ENABLE_ID_POOL_VALIDATION = "validateIdentificationPool"
         internal const val SINGLE_GOOD_QUALITY_FALLBACK_REQUIRED = "singleQualityFallbackRequired"
         internal const val FACE_AUTO_CAPTURE_ENABLED = "faceAutoCaptureEnabled"
+        internal const val FACE_AUTO_CAPTURE_IMAGING_DURATION_MILLIS = "faceAutoCaptureImagingDurationMillis"
+
+        const val FACE_AUTO_CAPTURE_IMAGING_DURATION_MILLIS_MIN = 1L
+        const val FACE_AUTO_CAPTURE_IMAGING_DURATION_MILLIS_DEFAULT = 3_000L
+        const val FACE_AUTO_CAPTURE_IMAGING_DURATION_MILLIS_MAX = 10_000L
```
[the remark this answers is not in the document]
alex-vt: Sure, updated to 1 minute - should be a fairly long time in real-world image-taking scenarios. If still not enough for extreme cases, let me know and we can push it further. A reasonable limit still should be there to prevent an accidentally wrong configuration (with a too large value) from disrupting the capture operations in production.

infra/config-sync/src/main/java/com/simprints/infra/config/sync/ConfigManager.kt
```diff
@@ -45,6 +45,9 @@ class ConfigManager @Inject constructor(
         }
     }
 
+    // Watches the local state of the project configuration without attempting to download if empty
```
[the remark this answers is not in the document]
alex-vt: Thank you. It was originally intended for pure local state observation, that is enough for the purpose. But sure, a download-if-empty makes it more robust, particularly if we rely on just `watchProjectConfiguration` and remove `getProjectConfiguration` in some places in the future. Updated.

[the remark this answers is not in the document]
BurningAXE: I fail to see the need for a flow of project configuration at all - when is a single getProjectConfiguration() insufficient?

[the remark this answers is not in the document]
alex-vt: If the auto-capture is enabled or disabled in the config, and the user syncs the config changes to the device by pressing "Update configuration", the "Auto-capture" option will appear/disappear right away - it's state is being watched. Otherwise user would need to click back and then again into settings.

Not having the flow: that's a minor UI inconvenience, but also a "lazy" approach when UI is allowed to be inconsistent when data updates.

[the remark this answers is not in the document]
BurningAXE: Ah, I see! Sorry, I didn't mean that it definitely needs to trigger download (we don't expect it to be empty anyway) - just didn't get the purpose initially.

BurningAXE (approved):

alexandr-simprints (approved):

PR #1062, opened 2025-01-16
CoSync: adding explicit class types to the tokenizable fields. Removing extra filtering

alexandr-simprints:
- From now on, tokenizable strings in `subjectActions` JSON specify the explicit class type: `TokenizableString.Raw` or `TokenizableString.Tokenized`. This is done by passing a serializer to the `GetEnrolmentCreationEventForSubjectUseCase`.
- Removing extra filtering from the CommCare data source. Sometimes, the tokenizable fields in the `query.*` and `event.payload.*` classes are incorrectly treated as tokenized/not tokenized, and therefore the value comparison returns false negatives. By removing the filter we slightly increase the validation pool but that should not affect the performance because the most resource-intensive operation is the JSON parsing, and the filtration is happening after the JSON is parsed.

### DIFF
--- a/feature/client-api/src/main/java/com/simprints/feature/clientapi/usecases/GetEnrolmentCreationEventForSubjectUseCase.kt
+++ b/feature/client-api/src/main/java/com/simprints/feature/clientapi/usecases/GetEnrolmentCreationEventForSubjectUseCase.kt
@@ -1,5 +1,9 @@
 package com.simprints.feature.clientapi.usecases
 
+import com.fasterxml.jackson.databind.module.SimpleModule
+import com.simprints.core.domain.tokenization.TokenizableString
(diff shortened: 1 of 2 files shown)


luhmirin-s (approved):

BurningAXE:
> Removing extra filtering from the CommCare data source. Sometimes, the tokenizable fields in the query.* and event.payload.* classes are incorrectly treated as tokenized/not tokenized, and therefore the value comparison returns false negatives.
I don't get this - why are tokenizable fields' subtypes treated incorrectly? 
By removing the query, you will be searching in the whole DB instead of the module or user subset which is incorrect behaviour. And for verification it would be even worse.

alexandr-simprints:
Closing it for now, as it seems that this will not be an issue once the MS-859 issue is resolved

PR #1068, opened 2025-01-21
MS-810 Manual log file exporting

luhmirin-s:
* Replaced Timber with Kermit to allow easy log writing to rotated files
* Configure the file log writer
* Add a zip archive export button to the troubleshooting screen
  * Zip file is protected by the same settings password (in rare case when password is not set, it will use project ID instead).


feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
```diff
@@ -0,0 +1,85 @@
...
+            val file: File,
+        ) : LogsExportResult()
+
+        data object InProgress : LogsExportResult()
+
+        data object NotStarted : LogsExportResult()
+    }
+
+    companion object {
+        private const val ARCHIVE_NAME = "log_archives/"
```
BurningAXE, line 83: Should this come from feature/troubleshooting/src/main/res/xml/log_archive_paths.xml

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/OverviewViewModel.kt
```diff
@@ -22,6 +26,7 @@ internal class OverviewViewModel @Inject constructor(
     private val collectLicenseStates: CollectLicenceStatesUseCase,
     private val collectNetworkInformation: CollectNetworkInformationUseCase,
     private val doServerPing: PingServerUseCase,
+    private val doExportLogs: ExportLogsUseCase,
```
BurningAXE, line 29: `exportLogs`?

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/OverviewViewModel.kt
```diff
@@ -48,6 +53,10 @@ internal class OverviewViewModel @Inject constructor(
         get() = _scannerState
     private val _scannerState = MutableLiveData("Connecting to scanner...")
 
+    val logsExport: LiveData<LogsExportResult>
```
BurningAXE, line 56: `logsExportResult`?

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        if (files.isNotEmpty()) {
+            val archiveFile = createArchiveFile(dateFormatter.format(Date()) + "_$deviceId")
+            val password = getPassword()
+            createZipArchiveWithPassword(archiveFile, files.toList(), password)
+
+            delay(2000L) // To look like it does something
+            emit(LogsExportResult.Success(deviceId, archiveFile))
+        } else {
+            emit(LogsExportResult.NotStarted)
```
BurningAXE, line 40: Wouldn't this make users to tap the button again and again thinking it's not working?

feature/troubleshooting/src/test/java/com/simprints/feature/troubleshooting/overview/OverviewViewModelTest.kt
```diff
@@ -101,4 +109,28 @@ class OverviewViewModelTest {
             PingResult.Success("message"),
         )
     }
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun `propagates server log export result`() = runTest {
```
BurningAXE, line 115: server? :)

infra/logging/src/main/java/com/simprints/infra/logging/writers/OptionalTagFormatter.kt
```diff
@@ -0,0 +1,19 @@
...
+internal class OptionalTagFormatter : MessageStringFormatter {
+    override fun formatTag(tag: Tag): String = "(${tag.tag})"
+
+    override fun formatMessage(
+        severity: Severity?,
+        tag: Tag?,
+        message: Message,
+    ): String {
+        if (tag?.tag.isNullOrBlank()) return message.message
+        return "${formatTag(tag)} ${message.message}"
```
BurningAXE, line 17: Previously it was `[TAG] Message` which IMO is cleaner

fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v1/MessageDispatcher.java
```diff
@@ -111,7 +111,7 @@ private Message sendRecv(@NonNull Message request, @NonNull MESSAGE_TYPE answerT
             try {
                 pendingRequest.lock.wait(timeOutMs);
             } catch (InterruptedException e) {
-                Simber.i("Message response timed out", e);
+                Simber.Companion.i("Message response timed out", e);
```
BurningAXE, line 114: Not very excited about this pattern

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -1,8 +1,8 @@
 package com.simprints.infra.logging
 
+import co.touchlab.kermit.Logger
```
BurningAXE, line 3: I kind of assumed that we would add something to Simber, not completely change the underlying library. Couldn't we just add a new Tree to Timber?

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -176,4 +108,105 @@ object Simber {
...
+
+    companion object {
+        const val DEFAULT_TAG = "Simber"
+        const val USER_PROPERTY_TAG = "zzUserPropertyTag"
+        private const val FIREBASE_ANALYTICS_MAX_MESSAGE_LENGTH = 100
+        private const val FIREBASE_ANALYTICS_MAX_TAG_LENGTH = 24
+        private const val FIREBASE_ANALYTICS_MAX_USER_TAG_LENGTH = 40
+        private val firebaseInvalidCharactersRegex = Regex("[^a-zA-Z0-9_]")
+
+        val INSTANCE = Simber()
```
BurningAXE, line 120: Not that the current setup is ideal but this feels even more awkward

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
luhmirin-s: They might, but this is better than leaving it in a broken state.

feature/troubleshooting/src/test/java/com/simprints/feature/troubleshooting/overview/OverviewViewModelTest.kt
luhmirin-s: Thanks gemini :D

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
luhmirin-s: I will add a "failed" state to briefly show different text.

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/OverviewViewModel.kt
luhmirin-s: It separates the "action" use cases from the data collection ones.

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
BurningAXE: Yeah, that's what I meant - maybe just show a toast "Nothing to export".

feature/troubleshooting/src/main/java/com/simprints/feature/troubleshooting/overview/usecase/ExportLogsUseCase.kt
luhmirin-s: I don't think that it is possible to retrieve the value from paths file, tbh. As long as the values match, it should work.

fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v1/MessageDispatcher.java
luhmirin-s: This is only forced in the java files (of which we have 1 or 2), in kotlin you can do `Simber.i()` directly.

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
luhmirin-s: We would need to rewrite the whole file writing and rotating logic ourselves, which could cause more issues and work.

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
luhmirin-s: Only because the initial Simber interface (specifically the tag application) was modelled after Timber and I did not want another project-wide refactoring for logs in this PR. 

If we move the tags to be parameters in Simber methods, all those companion hacks could be removed. I can do it in a separate PR, after this is merged.

infra/logging/src/main/java/com/simprints/infra/logging/writers/OptionalTagFormatter.kt
luhmirin-s: This is the same, just with a check that tag is present. I will make the brackets square, to keep it closer to the old approach.

alexandr-simprints (approved):

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
BurningAXE: Yeah, I expected that we would plug-in Kermit as one more `CrashReportingTree`. Appreciate that makes things much more complex.
How confident are you switching to Kermit won't cause any major issues in prod? Timber, for all it's flaws, has served us well for a long time.

luhmirin-s: It should be fine since Kermit is developed by one of the largest mobile dev agencies online and they would be directly interested in supporting it.

BurningAXE: Checking it's GitHub - looks well supported, indeed. Frequent releases and just 3 open issues.

BurningAXE (approved):

PR #1070, opened 2025-01-22
update debug version name

TristramN:
update the workflow to dynamically get the version name based on the type of build. This is needed because for debug version the version name is not available in the branch name. 

(I will squash on merge :-D)


luhmirin-s:
@TristramN would it make sense to incorporate #1069 into this as well?

TristramN:
This PR now includes https://github.com/Simprints/Android-Simprints-ID/pull/1069. Context here: https://simprints.slack.com/archives/GAW460V19/p1737538428571869

alexandr-simprints (approved):

PR #1071, opened 2025-01-22
MS-810 Simplify Simber interface

luhmirin-s:
Addressing concern about the added complexity in Simber implementation when moving to Kermit.

### DIFF
--- a/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
+++ b/infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
@@ -13,53 +13,132 @@ import javax.net.ssl.SSLProtocolException
  * library.
  * @see <a href="URL#https://github.com/JakeWharton/timber">Timber</a>
  */
-class Simber(
-    tag: String = DEFAULT_TAG,
-    private val logger: Logger = Logger(Logger.config, tag),
-) {
-    fun d(message: String) = logger.d(message, null)
+object Simber {
+    /**
+     * Use this for debugging purposes. If you want to print out a bunch of messages so you can log
+     * the exact flow of your program, use this. If you want to keep a log of variable values,
+     * use this.
+     * DEBUG: Is sent to Log.d
+     * STAGING: Is sent to Log.d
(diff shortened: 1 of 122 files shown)


infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
```diff
@@ -13,53 +13,151 @@ import javax.net.ssl.SSLProtocolException
...
+    fun d(message: String) = d(message, DEFAULT_TAG)
+
+    fun d(
+        message: String,
+        tag: CrashReportTag,
+    ) = d(message, tag?.name ?: DEFAULT_TAG)
+
+    fun d(
+        message: String,
+        tag: String,
```
BurningAXE, line 36: Can we do `tag: String = DEFAULT_TAG` and get rid of the single parameter functions + the elvis in callers of this one?

BurningAXE (approved):
That's better!

infra/logging/src/main/java/com/simprints/infra/logging/Simber.kt
luhmirin-s: That was my initial approach, but the compiler/analyzer gets confused since there will be ambiguous signatures. I want to preserve the option to pass either enum or string, but if those have default values, for the compiler those "tag-less" calls match both signatures.

There were couple options to fix it: 
* having a dedicated call without tag
* forcing tag everywhere
* passing null/empty explicitly when we don't want the tag

I chose the first option since it was the cleanest one, but I can be easily persuaded to change it :D

BurningAXE: > if those have default values, for the compiler those "tag-less" calls match both signatures

Hm, even if only the string-one has a default?
If not possible, I guess the current one is best as at least the complexity is in one place only instead of in all calls.

luhmirin-s: I have simplified it by one method. 👍🏻

PR #1087, opened 2025-02-10
[MS-847] enhanced 1 n progress indicator

alexandr-simprints:
We would like to clearly communicate the progress on the matching screen to the user. During the CoSync reading it might take some time to read the records, so the progress bar is now set to reflect the actual progress, and an additional `(24/999)` text is added to display the progress.
image
image

https://github.com/user-attachments/assets/4037629b-2cdd-4427-893b-aa4f4da5d108


feature/matcher/src/main/java/com/simprints/matcher/usecases/MatcherUseCase.kt
```diff
@@ -13,7 +13,8 @@ internal interface MatcherUseCase {
      */
     suspend operator fun invoke(
         matchParams: MatchParams,
-        onLoadingCandidates: () -> Unit = {},
+        onLoadingStarted: (totalCandidates: Int) -> Unit = {},
+        onCandidateLoaded: () -> Unit = {},
```
meladRaouf, line 17: regarding onLoadingStarted  and onCandidateLoaded  
Could the implementation be simplified using LiveData or StateFlow?

alexandr-simprints: It could, yes. I've just kept it similar to how it was implemented previously. Would you suggest moving to flow? Don't want to be dependent on android's LiveData within the use-case.

meladRaouf: Yes, Flow would be great.

luhmirin-s (approved):

meladRaouf (approved):

PR #1088, opened 2025-02-11
MS-871 Record update support

luhmirin-s:
* Replaced a single `PersonCreationEvent` with `BiometricReferenceCreationEvent` per biometric capture. 
* Updated the enrolment event to have the list of biometrics-referenced IDs instead of the person creation event ID in the payload. Additionally, added the probe reference ID to the 1:1 and 1:N match events.
* Added biometric reference ID to the local subject biometric sample records with a forced override/migration in the next sync cycle.
* Added support for subject update event in down-sync.

More details about the changes are available in the [design doc](https://docs.google.com/document/d/1fhCPZfpsgfhwJ4lez1GnohWJN-zCX7p_FgP7hkuw2V0/edit?tab=t.lrqvqxqdapwi)


infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
```diff
@@ -310,6 +311,7 @@ internal class EventUpSyncTask @Inject constructor(
 
         config.canSyncBiometricDataToSimprints() -> events.filter {
```
ybourgery, line 312: You are missing the `BiometricReferenceCreation` here as it's a biometric and analytics event

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
```diff
@@ -251,6 +257,18 @@ internal class EventDownSyncTask @Inject constructor(
...
 
+    private suspend fun handleSubjectUpdateEvent(event: EnrolmentRecordUpdateEvent): List<SubjectAction> {
+        val existingSubject = enrolmentRecordRepository.load(SubjectQuery(subjectId = event.payload.subjectId)).first()
+        val subject = subjectFactory.buildSubjectFromUpdatePayload(existingSubject, event.payload)
+
+        return if (subject.fingerprintSamples.isNotEmpty() || subject.faceSamples.isNotEmpty()) {
+            listOf(Write(subject))
+        } else {
+            // Having no samples after update is equal to deletion
+            listOf(Deletion(event.payload.subjectId))
```
ybourgery, line 268: If you delete it, and after (in a new down-sync request) you receive a new enrolment record update, will the line `enrolmentRecordRepository.load(SubjectQuery(subjectId = event.payload.subjectId)).first()` not fail with a `NoSuchElementException`?

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
```diff
@@ -251,6 +257,18 @@ internal class EventDownSyncTask @Inject constructor(
     private fun handleSubjectDeletionEvent(event: EnrolmentRecordDeletionEvent): List<SubjectAction> =
         listOf(Deletion(event.payload.subjectId))
 
+    private suspend fun handleSubjectUpdateEvent(event: EnrolmentRecordUpdateEvent): List<SubjectAction> {
+        val existingSubject = enrolmentRecordRepository.load(SubjectQuery(subjectId = event.payload.subjectId)).first()
```
ybourgery, line 261: If the record is created and updated in the same batch, my understanding of the code is that this line will fail as the record would not be yet, created as all the operations for the database of a given batch are applied at the end

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -179,8 +178,10 @@ internal class FaceCaptureViewModel @Inject constructor(
                     ),
                 )
             }
+            val referenceId = items.mapNotNull { it.sample?.template }.faceReferenceId().orEmpty()
```
ybourgery, line 181: Why do you need to generate a deterministic id instead of a random one here?

ybourgery (changes requested):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
luhmirin-s: Good point. It does feel like this could be an edge case when something went wrong, tho.

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
luhmirin-s: It was done that way before and I assumed there was a reason for it.

ybourgery: The only reason that I could think of is if you generate several times this reference id, but it doesn't seem so. What do you think about just generating a random id to keep it simpler?

luhmirin-s: I don't see much difference in the complexity here, it is a single call capture. So I would prefer keeping it until we are certain that nothing breaks if we change it to random UUID.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/EventDownSyncTask.kt
luhmirin-s: Moved the update logic fully into the local data source implementation.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/domain/models/SubjectAction.kt
```diff
@@ -1,13 +1,22 @@
 package com.simprints.infra.enrolment.records.store.domain.models
 
 import androidx.annotation.Keep
+import com.simprints.core.domain.face.FaceSample
+import com.simprints.core.domain.fingerprint.FingerprintSample
 
 @Keep
 sealed class SubjectAction {
-    data class Creation(
+    data class Write(
```
ybourgery, line 9: As you have moved the logic for the update, should you not rename this one back too?

feature/enrol-last-biometric/src/test/java/com/simprints/feature/enrollast/screen/EnrolLastBiometricViewModelTest.kt
```diff
@@ -213,30 +214,37 @@ internal class EnrolLastBiometricViewModelTest {
     }
 
     @Test
-    fun `Uses latest PersonCreationEvent for Enrolment event`() = runTest {
-        val personCreationEvent1 = mockk<PersonCreationEvent> {
+    fun `Uses all PersonCreationEvents for Enrolment event`() = runTest {
```
BurningAXE, line 217: `BiometricReferenceCreationEvent`s? :)

infra/events/src/test/java/com/simprints/infra/events/event/domain/validators/EnrolmentEventValidatorTest.kt
```diff
@@ -22,22 +23,22 @@ internal class EnrolmentEventValidatorTest {
     @Test
     fun validate_shouldValidateIfBiometricCaptureAndPersonCreationIsPresent() {
```
BurningAXE, line 24: AndBiometricReferenceCreationIsPresent?😇

infra/events/src/main/java/com/simprints/infra/events/event/domain/models/OneToOneMatchEvent.kt
```diff
@@ -22,40 +24,82 @@ data class OneToOneMatchEvent(
...
+            override val result: MatchEntry?,
+            override val fingerComparisonStrategy: FingerComparisonStrategy?,
+            val probeBiometricReferenceId: String,
+        ) : OneToOneMatchPayload(createdAt, eventVersion, endedAt, candidateId, matcher, result, fingerComparisonStrategy)
     }
 
     companion object {
-        const val EVENT_VERSION = 3
+        const val VERSION_WITHOUT_REFERENCE_ID = 3
+        const val EVENT_VERSION = 4
```
BurningAXE, line 103: `EVENT_VERSION` is confusing!?

feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/StartBackgroundSyncUseCase.kt
```diff
@@ -10,11 +10,10 @@ internal class StartBackgroundSyncUseCase @Inject constructor(
...
-        syncOrchestrator.scheduleBackgroundWork()
-
         val frequency = configManager.getProjectConfiguration().synchronization.frequency
-        if (frequency == SynchronizationConfiguration.Frequency.PERIODICALLY_AND_ON_SESSION_START) {
-            syncOrchestrator.startEventSync()
-        }
+
+        syncOrchestrator.scheduleBackgroundWork(
+            withDelay = frequency != SynchronizationConfiguration.Frequency.PERIODICALLY_AND_ON_SESSION_START,
+        )
```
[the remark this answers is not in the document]
luhmirin-s: As per documentation, the periodic worker is scheduled to run as soon as possible if the initial delay is not provided. This change removes the "race condition" when 2 concurrent syncs are executed at the start.

meladRaouf (approved):

infra/events/src/main/java/com/simprints/infra/events/event/domain/models/OneToOneMatchEvent.kt
```diff
@@ -22,40 +24,82 @@ data class OneToOneMatchEvent(
...
+    @JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.EXISTING_PROPERTY,
+        property = "eventVersion",
+        visible = true,
+    )
+    @JsonSubTypes(
+        JsonSubTypes.Type(
+            value = OneToOneMatchPayload.OneToOneMatchPayloadV3::class,
+            name = VERSION_WITHOUT_REFERENCE_ID.toString(),
```
BurningAXE, line 56: Probably a stupid question but in why do we need the old version?

feature/login-check/src/main/java/com/simprints/feature/logincheck/usecases/StartBackgroundSyncUseCase.kt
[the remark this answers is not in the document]
BurningAXE: Am I interpreting the new setup correctly that if frequency is just periodically and we have lots of sessions during the day, we'll keep pushing the start of the worker by 1 hour until there are no sessions for at least an hour?

infra/event-sync/src/main/java/com/simprints/infra/eventsync/status/down/domain/EventDownSyncOperation.kt
```diff
@@ -19,22 +19,9 @@ internal data class EventDownSyncOperation(
...
-                        modes.joinToString { it.name } +
-                        oldTypes
-                ).toByteArray(),
-            ).toString()
-    }
-
-    companion object {
-        // We need to keep this old types otherwise the unique key of the down-sync will change
-        // and we will need to down-sync again from scratch.
-        internal var oldTypes = "ENROLMENT_RECORD_CREATION, ENROLMENT_RECORD_MOVE, ENROLMENT_RECORD_DELETION"
```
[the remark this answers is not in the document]
luhmirin-s: Kinda, sneaky way to force resync is the DB version bump; the removal of deprecated stuff is just a nice side effect.

infra/event-sync/schemas/com.simprints.infra.eventsync.status.EventSyncStatusDatabase/4.json
```diff
@@ -0,0 +1,84 @@
...
+          ]
+        },
+        "indices": [],
+        "foreignKeys": []
+      }
+    ],
+    "views": [],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, 'fb9ad2ed04bc63d696f394aa7b2351c2')"
```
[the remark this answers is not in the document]
luhmirin-s: This is auto-generated by room due to the version bump.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/remote/models/fingerprint/ApiFingerprintReference.kt
```diff
@@ -17,7 +15,7 @@ internal data class ApiFingerprintReference(
 
 internal fun List<FingerprintSample>.toApi(encoder: EncodingUtils): ApiFingerprintReference? = if (isNotEmpty()) {
     ApiFingerprintReference(
-        UUID.nameUUIDFromBytes(concatTemplates()).toString(),
+        first().referenceId,
```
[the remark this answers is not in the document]
luhmirin-s: The reference ID should be the same for all templates within the capture.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/remote/models/face/ApiFaceReference.kt
```diff
@@ -17,10 +15,8 @@ internal data class ApiFaceReference(
 
 internal fun List<FaceSample>.toApi(encoder: EncodingUtils): ApiFaceReference? = if (isNotEmpty()) {
     ApiFaceReference(
-        UUID.nameUUIDFromBytes(concatTemplates()).toString(),
-        map {
-            ApiFaceTemplate(encoder.byteArrayToBase64(it.template))
-        },
+        first().referenceId,
```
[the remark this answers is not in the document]
luhmirin-s: The reference ID should be the same for all templates within the capture.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
```diff
@@ -129,23 +131,59 @@ internal class EnrolmentRecordLocalDataSourceImpl @Inject constructor(
...
-                        )
+                        val newSubject = action.subject
+                            .copy(
+                                moduleId = action.subject.moduleId.tokenizeIfNecessary(TokenKeyType.ModuleId, project),
+                                attendantId = action.subject.attendantId.tokenizeIfNecessary(TokenKeyType.AttendantId, project),
+                            ).fromDomainToDb()
+                        val dbSubject: DbSubject? = realm.findSubject(newSubject.subjectId)
+
+                        if (dbSubject != null) {
+                            // When updating an existing subject, we must manually delete outdated samples
```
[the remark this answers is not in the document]
luhmirin-s: In the context of enrolment DB, "Creation" is also the "Update", especially when we are silently re-syncing the records. Unfortunately, the local sample IDs are always randomly generated, so when we download the same subject to update the "referenceId" fields, it will always write new samples into the database.

infra/events/src/main/java/com/simprints/infra/events/event/domain/models/OneToOneMatchEvent.kt
luhmirin-s: We are storing those models as String in the event database, so the old version should still be un-marshallable even after SID upgrades. And migrating data in those JSON strings isn't really possible since we don't have the correct reference ID for each (at least within a reasonable effort).

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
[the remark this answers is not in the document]
BurningAXE: Hm, why is BFSID generating new referenceIds?

infra/events/src/main/java/com/simprints/infra/events/event/domain/models/OneToOneMatchEvent.kt
BurningAXE: Aren't we uploading the JSONs directly as strings? 
Or is the un-marshalling needed for the new diagnostic screen that shows events?

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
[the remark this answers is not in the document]
luhmirin-s: BFSID is tracking the reference IDs from the person creation events, SID was not until this update. So before this update, local record data did not have a reference ID, and there is no way to get them without the re-sync.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/SubjectFactory.kt
```diff
@@ -92,45 +111,46 @@ class SubjectFactory @Inject constructor(
...
 
     private fun extractFaceSamples(faceResponse: FaceCaptureResult) = faceResponse.results
         .mapNotNull { it.sample }
-        .map { FaceSample(it.template, it.format) }
+        .map { FaceSample(it.template, it.format, faceResponse.referenceId) }
 
-    private fun extractFingerprintSamplesFromBiometricReferences(biometricReferences: List<BiometricReference>?) = biometricReferences
+    fun extractFingerprintSamplesFromBiometricReferences(biometricReferences: List<BiometricReference>?) = biometricReferences
         ?.filterIsInstance<FingerprintReference>()
         ?.firstOrNull()
```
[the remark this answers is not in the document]
luhmirin-s: Good point. I updated the code to do a map instead.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/tasks/SubjectFactory.kt
```diff
@@ -92,45 +111,46 @@ class SubjectFactory @Inject constructor(
...
         template = encodingUtils.base64ToBytes(template.template),
         templateQualityScore = template.quality,
         format = format,
+        referenceId = referenceId,
     )
 
-    private fun extractFaceSamplesFromBiometricReferences(biometricReferences: List<BiometricReference>?) = biometricReferences
+    fun extractFaceSamplesFromBiometricReferences(biometricReferences: List<BiometricReference>?) = biometricReferences
         ?.filterIsInstance<FaceReference>()
         ?.firstOrNull()
```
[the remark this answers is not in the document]
luhmirin-s: Good point. I updated the code to do a map instead.

infra/events/src/main/java/com/simprints/infra/events/event/domain/models/OneToOneMatchEvent.kt
luhmirin-s: Not as string, there is a mapping from domain models to API models in between.

luhmirin-s: To be precise: from DB model string field to domain model to API model. :D

ybourgery (approved):

PR #1092, opened 2025-02-13
[MS-859] Using explicit tokenizable strings in api calls

alexandr-simprints:
## Overview
When sending events to the BFSID, we need to make sure that all the values necessary values are tokenized.

## Implementation
- `TokenizeEventPayloadFieldsUseCase` is added. This use case gets the tokenizable fields of each `Event` object and checks whether they are tokenized. If not, attempts to tokenize said fields.
- `Event.fromDomainToApi` now takes the abovementioned use-case as an argument, and runs the tokenization before mapping the domain Event model to the API model.

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
@@ -3,15 +3,9 @@ package com.simprints.infra.eventsync.event.remote.models.session
 import androidx.annotation.Keep
 import com.fasterxml.jackson.annotation.JsonInclude
 import com.fasterxml.jackson.annotation.JsonInclude.Include
-import com.simprints.infra.config.store.models.Project
-import com.simprints.infra.config.store.tokenization.TokenizationProcessor
-import com.simprints.infra.events.event.domain.models.Event
-import com.simprints.infra.events.event.domain.models.scope.EventScope
 import com.simprints.infra.eventsync.event.remote.models.ApiEvent
 import com.simprints.infra.eventsync.event.remote.models.ApiModality
 import com.simprints.infra.eventsync.event.remote.models.ApiTimestamp
-import com.simprints.infra.eventsync.event.remote.models.fromDomainToApi
-import com.simprints.infra.eventsync.event.usecases.TokenizeEventPayloadFieldsUseCase
 
 @Keep
 internal data class ApiEventScope(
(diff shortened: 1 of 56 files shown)


infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
```diff
@@ -48,7 +53,7 @@ internal data class ApiEventScope(
             location = scope.payload.location?.fromDomainToApi(),
             projectConfigurationUpdatedAt = scope.payload.projectConfigurationUpdatedAt,
             projectConfigurationId = scope.payload.projectConfigurationId.orEmpty(),
-            events = events.map { it.fromDomainToApi() },
+            events = events.map { it.fromDomainToApi(tokenizeEventPayloadFieldsUseCase, project) },
```
luhmirin-s, line 56: Not a fan of providing a use case as a parameter to a static function, would it be possible to make the the whole `fromDomainToApi` function into a `MapDomainEventToApiUseCase` instead? And probably do the same for `MapDomainEventScopeToApiUseCase`.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/usecases/TokenizeEventPayloadFieldsUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.simprints.infra.config.store.models.Project
+import com.simprints.infra.config.store.tokenization.TokenizationProcessor
+import javax.inject.Inject
+import com.simprints.infra.events.event.domain.models.Event
+
+internal class TokenizeEventPayloadFieldsUseCase @Inject constructor(
+    private val tokenizationProcessor: TokenizationProcessor
+) {
+    operator fun invoke(event: Event, project: Project): Event {
+        val tokenizedFields = event.getTokenizedFields().mapValues { entry ->
```
BurningAXE, line 13: Naming is a little confusing here, first we `getTokenizedFields()` and then check whether they are indeed tokenized and do tokenize some of them :D Would it be more appropriate to rename Event function to `getTokenizableFields()`? After all it returns the parent `TokenizableString`

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
```diff
@@ -87,7 +93,7 @@ internal class EventUpSyncTask @Inject constructor(
                 createUpSyncContentContent = {
                     isUsefulUpload = it > 0
                     EventUpSyncRequestEvent.UpSyncContent(sessionCount = it)
-                },
+                }
```
BurningAXE, line 96: Hmm, is this KtLint?

luhmirin-s: KtLint would do the opposite 🤔

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/session/ApiEventScope.kt
alexandr-simprints: Agree, that makes it easier to test, and presents code in a more flexible way to treat business logic in mapping between layers. Please see the added `MapDomainEventsToApiUseCase` in the latest commits

alexandr-simprints:
Please review the latest changes, and ignore the test coverage, as they are due to refactoring in the `*Event.kt` model files

luhmirin-s (approved):
Is the coverage so low purely due to the renaming of methods?

PR #1093, opened 2025-02-13
[MS-852] Tokenizable fields from CommCare response are now properly filtered

alexandr-simprints:
## Overview
Given that the plain string values might be both tokenized or untokenized, we need some mechanism that properly compares the values in the `CommCareIdentityDataSource`.

## Solutions
The `CompareImplicitTokenizedStringsUseCase` is added to check is the implicitly tokenized values are the same. Implicitly tokenized means that the plain string value can be both tokenized and untokenized, and we have no additional information that specifies the tokenization state. 

The use case operates using the following logic:
- Checks if the values are already similar
- If not, then tries to tokenize both values so that each of them is wrapped into `TokenizableString.Tokenized` class. It tries to decrypt each string using the project tokenization keys. If successful, then the string is already tokenized. If not, it tokenizes the value. Once both strings are tokenized, it compares their resulting value.

### DIFF
--- a/infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
+++ b/infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
@@ -13,11 +13,14 @@ import com.simprints.core.domain.tokenization.serialization.TokenizationClassNam
 import com.simprints.core.domain.tokenization.serialization.TokenizationClassNameSerializer
 import com.simprints.core.tools.json.JsonHelper
 import com.simprints.core.tools.utils.EncodingUtils
+import com.simprints.infra.config.store.models.Project
+import com.simprints.infra.config.store.models.TokenKeyType
 import com.simprints.infra.enrolment.records.store.IdentityDataSource
 import com.simprints.infra.enrolment.records.store.domain.models.BiometricDataSource
 import com.simprints.infra.enrolment.records.store.domain.models.FaceIdentity
 import com.simprints.infra.enrolment.records.store.domain.models.FingerprintIdentity
 import com.simprints.infra.enrolment.records.store.domain.models.SubjectQuery
+import com.simprints.infra.enrolment.records.store.usecases.CompareImplicitTokenizedStringsUseCase
 import com.simprints.infra.events.event.cosync.CoSyncEnrolmentRecordEvents
 import com.simprints.infra.events.event.domain.models.subject.EnrolmentRecordCreationEvent
 import com.simprints.infra.events.event.domain.models.subject.FaceReference
@@ -28,10 +31,12 @@ import dagger.hilt.android.qualifiers.ApplicationContext
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.withContext
 import org.json.JSONException
+import javax.inject.Inject
 
-internal class CommCareIdentityDataSource(
+internal class CommCareIdentityDataSource @Inject constructor(
     private val encoder: EncodingUtils,
     private val jsonHelper: JsonHelper,
+    private val compareImplicitTokenizedStringsUseCase: CompareImplicitTokenizedStringsUseCase,
     @ApplicationContext private val context: Context,
     @DispatcherIO private val dispatcher: CoroutineDispatcher,
 ) : IdentityDataSource {
@@ -51,9 +56,10 @@ internal class CommCareIdentityDataSource(
         query: SubjectQuery,
         range: IntRange,
         dataSource: BiometricDataSource,
+        project: Project,
         onCandidateLoaded: () -> Unit,
     ): List<FingerprintIdentity> = withContext(dispatcher) {
-        loadEnrolmentRecordCreationEvents(range, dataSource.callerPackageName(), query, onCandidateLoaded)
+        loadEnrolmentRecordCreationEvents(range, dataSource.callerPackageName(), query, project, onCandidateLoaded)
             .filter { erce ->
                 erce.payload.biometricReferences.any { it is FingerprintReference && it.format == query.fingerprintSampleFormat }
             }.map {
@@ -79,13 +85,14 @@ internal class CommCareIdentityDataSource(
         range: IntRange,
         callerPackageName: String,
         query: SubjectQuery,
+        project: Project,
         onCandidateLoaded: () -> Unit,
     ): List<EnrolmentRecordCreationEvent> {
         val enrolmentRecordCreationEvents: MutableList<EnrolmentRecordCreationEvent> = mutableListOf()
         try {
             val caseId = attemptExtractingCaseId(query.metadata)
             if (caseId != null) {
-                return loadEnrolmentRecordCreationEvents(caseId, callerPackageName, query)
+                return loadEnrolmentRecordCreationEvents(caseId, callerPackageName, query, project)
             }
 
             context.contentResolver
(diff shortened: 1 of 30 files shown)


infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/usecases/CompareTokenizedStringsUseCase.kt
```diff
@@ -0,0 +1,47 @@
...
+ *
+ *      Even though plain string values are different, they represent the same entity. s1 is going to be encrypted and compared to the s2.
+ */
+class CompareImplicitTokenizedStringsUseCase @Inject constructor(
+    private val tokenizationProcessor: TokenizationProcessor
+) {
+    operator fun invoke(s1: String?, s2: String, tokenKeyType: TokenKeyType, project: Project): Boolean = when {
+        s1 == null -> false
+        s1 == s2 -> true
+        else -> tokenize(s1, tokenKeyType, project) == tokenize(s2, tokenKeyType, project)
```
BurningAXE, line 27: Aren't we unsure about the tokenization state of the event's moduleId/attendantId only? 
By testing the tokenization state of the query's moduleId/attendantId, we'll be adding a needless decryption step multiplied by the number of records.....to an already very slow process.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -76,14 +81,15 @@ internal class CommCareIdentityDataSource @Inject constructor(
         range: IntRange,
         callerPackageName: String,
         query: SubjectQuery,
+        project: Project,
```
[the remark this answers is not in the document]
alexandr-simprints: Honestly, I don't really like it either.

However, after spending hours resolving circular dependencies, and refactoring different config sources and repositories, I'd like to keep as few class dependencies as possible. The only way I see it happening, if we add the direct `DataStore<ProtoProjectConfiguration>` dependency to avoid one data store to rely on another (i.e. `ConfigLocalDataSource`).

Would we be happy with it or does any other approach come to mind?

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/usecases/CompareTokenizedStringsUseCase.kt
alexandr-simprints: ### Historical context
Due to mistake in the tokenizable string serialization made in the previous versions of SID, we cannot be sure if the `event.payload.moduleId` is tokenized or not. Some values returned from the eCHIS database are encrypted but they have no indication of tokenization, i.e. they are saved as `{ "value": "tokenized-value" }` instead of `{ "className": "TokenizableString.Tokenized", "value": "tokenized-value" }`.
When such values are deserialized, they become `TokenizableString.Raw("tokenized-value")` because no className is present (see `TokenizationClassNameDeserializer`).


### Current solution
Both `SubjectQuery.moduleId` and `SubjectQuery.attendantId` are stored as `String`. We need to know whether the plain string is tokenized or not.
Since we are also cannot be sure about the tokenization state of the fields in the `Event` after deserialization, we also need to check whether its value is encrypted or not. Let's take a look at the example:
 - First argument, `query.attendantId` - `String` value that can be both tokenized or untokenized. Needs to be checked explicitly by trying to encrypt the value. If decryption is successful, then no more actions are taken. If decryption failed, assuming it is untokenized value, and tokenize it. Maximum amount of actions: 2 (_decyprt + encrypt_).
 - Second argument, `event.payload.attendantId` - `TokenizableString` that can be both `TokenizableString.Tokenized` and `TokenizableString.Raw`. But we know that in some cases the `TokenizableString.Raw` class might actually have tokenized value, and therefore the only option is to explicitly check its value for tokenization the same way we did in the previous step. Maximum amount of actions: 2 (_decyprt + encrypt_); minimum: 1 (_decrypt_)

The only way to optimize the algorithm is to pass `s2: TokenizableString` and perform no action to it if it is already encrypted `TokenizableString.Tokenized`. This string will only be tokenized if the filed was saved as `{ "className": "TokenizableString.Tokenized", "value": "tokenized-value" }` in JSON, therefore we can save an extra step for such cases
```
    operator fun invoke(s1: String?, s2: TokenizableString, tokenKeyType: TokenKeyType, project: Project): Boolean = when {
        s1 == null -> false
        s1 == s2.value -> true
  
        // Adding this step
        s2 is TokenizableString.Tokenized -> tokenize(s1, tokenKeyType, project) == s2

        else -> tokenize(s1, tokenKeyType, project) == tokenize(s2, tokenKeyType, project)
```

BurningAXE: > First argument, query.attendantId - String value that can be both tokenized or untokenized. Needs to be checked explicitly by trying to encrypt the value. If decryption is successful, then no more actions are taken. If decryption failed, assuming it is untokenized value, and tokenize it. Maximum amount of actions: 2 (decrypt + encrypt).

My point is that query.attendantId (and query.moduleId) are always `TokenizableString.Raw`. [We get](https://github.com/Simprints/Android-Simprints-ID/blob/c9f92e406153a084a4b18d4168294d0758b68e23/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/usecases/steps/BuildMatcherSubjectQueryUseCase.kt#L23) the SubjectQuery values from the ActionRequest:
`attendantId = actionRequest.userId.value`
and the [ActionRequest value is taken](https://github.com/Simprints/Android-Simprints-ID/blob/c9f92e406153a084a4b18d4168294d0758b68e23/feature/client-api/src/main/java/com/simprints/feature/clientapi/mappers/request/builders/IdentifyRequestBuilder.kt#L42) as TokenizableString.Raw:
`userId = extractor.getUserId().asTokenizableRaw()`

alexandr-simprints: Oh, I see your point, thanks!
My only issue with this is that we are only treating it as `Tokenizable.Raw` _now_, and if something changes on the way from extractor to the use case in future, there will be no way to spot it. This will make the entire filtering system broken, and will be more prone to human error since the types are not strictly controlled. I'd prefer not to risk it.

BurningAXE: This concerns me, too. But I'd argue that instead of checking which type it is, we should change it's type to be explicitly `TokenizableString.Raw`. Overall we should limit direct use of `TokenizableString` only to case where it's unavoidable exactly to avoid such issues.

alexandr-simprints: Sorry the last message left me a bit confused. Are you saying that we should change the signature of the use case function to `invoke(s1: TokenizableString.Raw?, s2: TokenizableString, [...])` or that we should change the `SubjectQuery.attendatId` and `.moduleId` to `TokenizableString`?

BurningAXE: Sorry, meant that the proper way to avoid errors coming from the way `SubjectQuery.attendantId/.moduleId` are treated is to change their type to `TokenizableString.Raw` - this will avoid any doubts and enforce it at the compiler level and won't degrade match performance.

alexandr-simprints: Unfortunately, this is exactly the scope of works that I've tried to avoid with this PR. I've already implemented such an approach in another `MS-852` branch that had the changes described in the recent Tokenization design document.
Let me run the performance comparison of the current solution vs. the previous one, and come back with the results.

BurningAXE: Oh, I have missed this! One more point for why we should clear this mess and have specific types!
That said, we can have it as technical debt and fix in a separate ticket.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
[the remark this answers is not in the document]
luhmirin-s: If it causes dependency hell, then passing the project is fine. We do need to address the dependencies in a more broad refactoring soon(tm).

luhmirin-s (approved):

meladRaouf (approved):

PR #1094, opened 2025-02-13
[MS-721] Investigate what cause first enrolment to be unidentifiable on this specific project

alexandr-simprints:
## Overview
If the enrolment callout is made when there is no authentication in SID, then the `moduleId` and `attendantId` remain untokenized in the local database. Any further verification and identification request will have their `moduleId` tokenized, meaning that the initial enrolment will never be fetched from the local database, since its `moduleId` is untokenized.

## Example
- No SID auth
- Enrolment request is sent to SID with `moduleId` = `123`
- No tokenization keys are available at the moment of the event and subject creation, thus all the data from the request is saved with `moduleId` = `123`
- SID authenticates, tokenization keys become available, enrolment flow ends, data is saved in the local database
- Identification request is sent to SID with `moduleId` = `123`
- Since the tokenization keys are now available, `moduleId` becomes `encrypted_123`
- SID fetches all the subjects from module `encrypted_123`, receives 0 candidates, and returns 0 matches to the callout app

### `EnrolmentRecordLocalDataSource`
When saving subject using the `performActions` method, the tokenization keys should already be available. Trying to reactively tokenize all tokenizable fields before saving
```
is SubjectAction.Creation -> {
    val tokenizedModuleId = moduleId.tokenizeIfNecessary()
    val tokenizedAttendantId = attendantId.tokenizeIfNecessary()
    realm.copyToRealm(
        action.subject.copy(moduleId = tokenizedModuleId, attendantId = tokenizedAttendantId)
    )
}
```
## Reactively tokenizing modules
The goal is to provide the tokenized `moduleId` values to all callers that work with modules.

### `ConfigRepository`
When fetching the project configuration, reactively tokenizing modules before returning the config to caller.
```
fun getProjectConfiguration() {
    val config = localDataSource.getProjectConfiguration()
    return tokenizeModules(config)
}

fun watchProjectConfiguration() = localDataSource.watchProjectConfiguration().map { config ->
    tokenizeModules(config)
}
```
### `ConfigLocalDataSource`
When fetching the device configuration, reactively tokenizing modules before returning the config to caller.
```
fun getDeviceConfiguration() {
    val config = deviceConfigDataStore.data.first().toDomain()
    val tokenizedModules = config.selectedModules.map { moduleId ->
        when(moduleId) {
            is TokenizableString.Raw -> tokenizationProcessor.encrypt(
                decrypted = moduleId,
                tokenKeyType = TokenKeyType.ModuleId,
                project = getProject()
            )
            is TokenizableString.Tokenized -> moduleId
        }
    }
    config.selectedModules = tokenizedModules
    return config
}
```

### DIFF
--- a/infra/config-store/src/main/java/com/simprints/infra/config/store/ConfigRepositoryImpl.kt
+++ b/infra/config-store/src/main/java/com/simprints/infra/config/store/ConfigRepositoryImpl.kt
@@ -2,6 +2,7 @@ package com.simprints.infra.config.store
 
 import androidx.annotation.VisibleForTesting
 import com.simprints.core.DeviceID
+import com.simprints.core.domain.tokenization.TokenizableString
 import com.simprints.infra.config.store.local.ConfigLocalDataSource
 import com.simprints.infra.config.store.models.DeviceConfiguration
 import com.simprints.infra.config.store.models.DeviceState
@@ -13,19 +14,23 @@ import com.simprints.infra.config.store.models.PrivacyNoticeResult.Succeed
 import com.simprints.infra.config.store.models.Project
 import com.simprints.infra.config.store.models.ProjectConfiguration
 import com.simprints.infra.config.store.models.ProjectWithConfig
+import com.simprints.infra.config.store.models.TokenKeyType
 import com.simprints.infra.config.store.remote.ConfigRemoteDataSource
+import com.simprints.infra.config.store.tokenization.TokenizationProcessor
 import com.simprints.infra.logging.Simber
 import com.simprints.infra.network.SimNetwork
 import com.simprints.infra.network.exceptions.BackendMaintenanceException
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
 import javax.inject.Inject
 
 internal class ConfigRepositoryImpl @Inject constructor(
     private val localDataSource: ConfigLocalDataSource,
     private val remoteDataSource: ConfigRemoteDataSource,
     private val simNetwork: SimNetwork,
+    private val tokenizationProcessor: TokenizationProcessor,
     @DeviceID private val deviceId: String,
 ) : ConfigRepository {
     companion object {
@@ -46,9 +51,14 @@ internal class ConfigRepositoryImpl @Inject constructor(
             }
         }
 
-    override suspend fun getProjectConfiguration(): ProjectConfiguration = localDataSource.getProjectConfiguration()
+    override suspend fun getProjectConfiguration(): ProjectConfiguration {
+        val config = localDataSource.getProjectConfiguration()
+        return tokenizeModules(config)
+    }
 
-    override fun watchProjectConfiguration(): Flow<ProjectConfiguration> = localDataSource.watchProjectConfiguration()
+    override fun watchProjectConfiguration(): Flow<ProjectConfiguration> = localDataSource.watchProjectConfiguration().map { config ->
+        tokenizeModules(config)
+    }
 
     override suspend fun getDeviceState(): DeviceState {
         val projectId = localDataSource.getProject().id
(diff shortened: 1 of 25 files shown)


luhmirin-s (approved):

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
```diff
@@ -138,6 +147,15 @@ internal class EnrolmentRecordLocalDataSourceImpl @Inject constructor(
         }
     }
 
+    private fun TokenizableString.tokenizeIfNecessary(tokenKeyType: TokenKeyType, project: Project) = when(this) {
```
BurningAXE, line 150: Maybe it's worth extracting this to TokenizableString to make it accessible everywhere.

infra/config-store/src/main/java/com/simprints/infra/config/store/ConfigRepositoryImpl.kt
```diff
@@ -81,6 +91,28 @@ internal class ConfigRepositoryImpl @Inject constructor(
         }
     }
 
+    private suspend fun tokenizeModules(config: ProjectConfiguration): ProjectConfiguration {
```
BurningAXE, line 94: Is this really needed? [Looks like](https://github.com/Simprints/Android-Simprints-ID/blob/1b8c10c6e84da68ee41f30ee2b5c10c4710ff222/infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiSynchronizationConfiguration.kt#L106) we get the module options already tokenized from the API. The only option I can see is for the device to have just upgraded from a pre-tokenization version and not yet synced new configuration. But in this case it would not have the keys to encrypt modules, anyway.

infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
```diff
@@ -76,7 +80,21 @@ internal class ConfigLocalDataSourceImpl @Inject constructor(
         configDataStore.updateData { it.toBuilder().clear().build() }
     }
 
-    override suspend fun getDeviceConfiguration(): DeviceConfiguration = deviceConfigDataStore.data.first().toDomain()
+    override suspend fun getDeviceConfiguration(): DeviceConfiguration {
```
BurningAXE, line 83: This works and the overhead in cases where modules are already tokenized (increasing majority) the overhead would probably be super small. Still, it feels wrong to do this check every time instead of just migrate saved modules once to tokenized ones.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
luhmirin-s: From the top of my head, the `project` argument might cause dependency issues if extracted into the "tokenization" module.

infra/config-store/src/main/java/com/simprints/infra/config/store/local/ConfigLocalDataSourceImpl.kt
luhmirin-s: I agree that a migration would be cleaner, on the other hand, this would ensure 100% consistency in all cases. 
And considering the amount of selected modules is usually in the 1-10 range, I would say the overhead is negligible.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/local/EnrolmentRecordLocalDataSourceImpl.kt
alexandr-simprints: > Maybe it's worth extracting this to TokenizableString to make it accessible everywhere.

I'd like to do that but unfortunately, as Sergejs mentioned above, this might lead to the circular dependency issues in the future

PR #1103, opened 2025-02-18
Use the correct GOOGLE_SERVICES_FILE env var

meladRaouf:


meladRaouf:
Not the correct solution to the problem

PR #1109, opened 2025-02-19
Fix release pipeline

meladRaouf:


alex-vt (approved):

luhmirin-s (approved):
Maybe it would be better to merge this into the main first and then start a real release branch from it?

meladRaouf:
> Maybe it would be better to merge this into the main first and then start a real release branch from it?

Yes, this would be a better approach, but I merged it :D

PR #1113, opened 2025-02-24
[MS-887] Fix Filtering Logic in CommCareIdentityDataSource 

meladRaouf:
There was a bug in the identification callouts from CommCare due to incorrect filtering logic in CommCareIdentityDataSource. The filtering logic was unintentionally excluding all records, causing identification failures.

The issue arose when building the subjectQuery in BuildMatcherSubjectQueryUseCase, where the filtering logic failed under different partitionType conditions:

    partitionType = PROJECT → Both module ID and attendant ID are null.
    partitionType = USER → The module ID in the query is null.
    partitionType = MODULE → The attendant ID is null in the query.

Due to this, the filtering logic always failed, as it did not account for null values correctly.

I updated the filtering logic to correctly handle null values to ensure that valid records are no longer filtered out incorrectly.

### DIFF
--- a/infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
+++ b/infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
@@ -169,25 +169,44 @@ internal class CommCareIdentityDataSource @Inject constructor(
                 coSyncEnrolmentRecordEvents
                     ?.events
                     ?.filterIsInstance<EnrolmentRecordCreationEvent>()
-                    ?.filterNot { event ->
+                    ?.filter { event ->
                         // [MS-852] Plain strings from CommCare might be tokenized or untokenized. The only way to properly compare them
                         // is by trying to decrypt the values to check if already tokenized, and then compare the values
-                        (query.subjectId != null && query.subjectId != event.payload.subjectId) ||
-                            !compareImplicitTokenizedStringsUseCase(
-                                query.attendantId,
-                                event.payload.attendantId,
-                                TokenKeyType.AttendantId,
-                                project,
-                            ) ||
-                            !compareImplicitTokenizedStringsUseCase(
-                                query.moduleId,
-                                event.payload.moduleId,
-                                TokenKeyType.ModuleId,
-                                project,
-                            )
+                        isSubjectIdNullOrMatching(query, event) &&
+                            isAttendantIdNullOrMatching(query, event, project) &&
+                            isModuleIdNullOrMatching(query, event, project)
                     }
             }.orEmpty()
     }
+
+    private fun isSubjectIdNullOrMatching(
+        query: SubjectQuery,
+        event: EnrolmentRecordCreationEvent,
+    ): Boolean = query.subjectId == null || query.subjectId == event.payload.subjectId
+
+    private fun isAttendantIdNullOrMatching(
+        query: SubjectQuery,
+        event: EnrolmentRecordCreationEvent,
+        project: Project,
+    ): Boolean = query.attendantId == null ||
+        compareImplicitTokenizedStringsUseCase(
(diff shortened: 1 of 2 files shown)


infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -169,26 +169,43 @@ internal class CommCareIdentityDataSource @Inject constructor(
...
-                                TokenKeyType.AttendantId,
-                                project,
-                            ) ||
-                            !compareImplicitTokenizedStringsUseCase(
-                                query.moduleId,
-                                event.payload.moduleId,
-                                TokenKeyType.ModuleId,
-                                project,
-                            )
+                        isSubjectIdMatching(query, event) &&
```
luhmirin-s, line 175: This line is very counter-intuitive since it reads as "filter items that match by all 3 fields". 
I get the logic behind it, but the naming should be clearer, maybe at least something like "is<field>NullOrMatching".

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/usecases/CompareImplicitTokenizedStringsUseCase.kt
```diff
@@ -30,7 +30,7 @@ class CompareImplicitTokenizedStringsUseCase @Inject constructor(
         tokenKeyType: TokenKeyType,
         project: Project,
     ): Boolean = when {
-        s1 == null -> false
+        s1 == null -> true // ignore null values as they are not tokenized
```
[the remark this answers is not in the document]
meladRaouf: Yes, you are right, I reverted it back to false.

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
meladRaouf: done

infra/enrolment-records-store/src/main/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSource.kt
```diff
@@ -169,26 +169,45 @@ internal class CommCareIdentityDataSource @Inject constructor(
...
-                                project,
-                            )
+                        isSubjectIdNullOrMatching(query, event) &&
+                            isAttendantIdNullOrMatching(query, event, project) &&
+                            isModuleIdNullOrMatching(query, event, project)
                     }
             }.orEmpty()
     }
 
+    fun isSubjectIdNullOrMatching(
```
alexandr-simprints, line 182: Do the functions need to be public?

alexandr-simprints (approved):

infra/enrolment-records-store/src/test/java/com/simprints/infra/enrolment/records/store/commcare/CommCareIdentityDataSourceTest.kt
```diff
@@ -259,7 +260,14 @@ class CommCareIdentityDataSourceTest {
             SUBJECT_ACTIONS_FACE_2,
         )
         val templateFormat = "ROC_1_23"
-        val query = SubjectQuery(faceSampleFormat = templateFormat)
+        val query = SubjectQuery(
```
BurningAXE, line 263: Are the tests failing with the old implementation now?

meladRaouf: No Tests were OK.
I just added more parameters to one of the queries so that the tests covers all the flows in the filtering method

PR #1114, opened 2025-02-24
MS-885 Do not add verification success flag when threshold is not enabled

luhmirin-s:


gradle/libs.versions.toml
```diff
@@ -63,7 +63,7 @@ circleImageView_version = "3.1.0"
 kermit_version = "2.0.5"
 zip4j_version = "2.11.5"
 
-libsimprints_version = "2025.1.2-SNAPSHOT"
+libsimprints_version = "2025.1.3-SNAPSHOT"
```
meladRaouf, line 66: Are we going to release SID using a snapshot libsimprints?

luhmirin-s: It seems so. At least the initial release, as per the Slack discussion.

meladRaouf (approved):

PR #1116, opened 2025-02-24
[MS-894] Add a custom CoSync deserializer that accounts for past versions of EnrolmentRecordCreationEvent

BurningAXE:
When using CoSync as data source SID fails to parse enrolments made with version < 2023.3.3 (first version that introduces tokenization). This is because in older versions `moduleId` and `attendantId` in `EnrolmentRecordCreationPayload` were plain `String`s instead of `TokenizableString`s.

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEnrolmentRecordCreationEventDeserializer.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEnrolmentRecordCreationEventDeserializer.kt
@@ -1,0 +1,61 @@
+package com.simprints.infra.events.event.cosync
+
+import com.fasterxml.jackson.core.JsonParser
+import com.fasterxml.jackson.databind.DeserializationContext
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer
+import com.simprints.core.domain.tokenization.TokenizableString
+import com.simprints.infra.events.event.domain.models.subject.BiometricReference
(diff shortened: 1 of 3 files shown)


meladRaouf:
@BurningAXE is this manadatory ? do we have any project that uses cosync and SID < 2023.3.3 ?

BurningAXE:
> @BurningAXE is this manadatory ? do we have any project that uses cosync and SID < 2023.3.3 ?

The question is do we have any project that has saved CoSync enrolments from SID < 2023.3.3? And the answer is unfortunately yes!

luhmirin-s (approved):

infra/events/src/main/java/com/simprints/infra/events/event/cosync/CoSyncEnrolmentRecordCreationEventDeserializer.kt
```diff
@@ -0,0 +1,61 @@
...
+        val payload = node["payload"]
+
+        val subjectId = payload["subjectId"].asText()
+        val projectId = payload["projectId"].asText()
+
+        // Try to parse as TokenizableString first, fall back to plain String
+        val moduleId = try {
+            ctxt.readTreeAsValue(payload["moduleId"], TokenizableString::class.java)
+        } catch (_: Exception) {
+            TokenizableString.Raw(payload["moduleId"].asText())
```
luhmirin-s, line 35: I assume it would not be possible to see if the value is tokenized just without the appropriate meta-data, right?

meladRaouf (approved):

alexandr-simprints (approved):

PR #1128, opened 2025-03-05
[MS-929] Fix 1:N pool count for CoSync

BurningAXE:
[MS-929](https://simprints.atlassian.net/browse/MS-929)

**Problem**
Both the fingerprint and face matcher get the pool size (recorded in OneToManyMatch.MatchPool.count) by calling:
`val totalCandidates = enrolmentRecordRepository.count(queryWithSupportedFormat, dataSource = matchParams.biometricDataSource)`
This is fine when using the local DB. However, when using CommCare as the biometric data source, we cannot use any of the query’s filtering criteria so instead we just count the whole ContentProvider URI. This means we get the count of all cases in CommCare, regardless of their module or even whether they have biometric data attached or not. This might often lead to incorrect pool size reported in the `OneToManyMatch` event.

**Fix**
The fix contains two parts:
1. In the matchers count the number of actually loaded candidates and report that upstream so it ends in the `OneToManyMatch` event
2. In CommCareIdentityProvider move the onCandidateLoaded callback after filtering so that it only count actually loaded candidates (similar to how the local DB works)

[MS-929]: https://simprints.atlassian.net/browse/MS-929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
+++ b/feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
@@ -45,19 +45,23 @@ internal class FaceMatcherUseCase @Inject constructor(
         val queryWithSupportedFormat = matchParams.queryForCandidates.copy(
             faceSampleFormat = faceMatcher.supportedTemplateFormat,
         )
-        val totalCandidates = enrolmentRecordRepository.count(
+        val expectedCandidates = enrolmentRecordRepository.count(
             queryWithSupportedFormat,
             dataSource = matchParams.biometricDataSource,
         )
-        if (totalCandidates == 0) {
+        if (expectedCandidates == 0) {
             send(MatcherState.Success(emptyList(), 0, faceMatcher.matcherName))
             return@channelFlow
         }
 
         Simber.i("Matching candidates", tag = crashReportTag)
-        send(MatcherState.LoadingStarted(totalCandidates))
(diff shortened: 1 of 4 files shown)


feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -76,9 +78,9 @@ internal class FaceMatcherUseCase @Inject constructor(
                 .toList()
         }
 
-        Simber.i("Matched $totalCandidates candidates", tag = crashReportTag)
+        Simber.i("Matched $loadedCandidates candidates", tag = crashReportTag)
 
-        send(MatcherState.Success(resultItems, totalCandidates, faceMatcher.matcherName))
+        send(MatcherState.Success(resultItems, loadedCandidates, faceMatcher.matcherName))
```
meladRaouf, line 83: I don't understand the difference between totalCandidates and loadedCandidates
this use case  sends  `send(MatcherState.LoadingStarted(totalCandidates)) `
and 
` send(MatcherState.Success(resultItems, loadedCandidates, faceMatcher.matcherName))`

luhmirin-s (approved):

meladRaouf (approved):

PR #1131, opened 2025-03-08
improve OkHttpClient reuse & efficiency

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-934)
Will be released in:  2025.2.0

First known affected version:From the moment we started using OkHttp 

The issue was identified through [this](https://console.firebase.google.com/project/simprints-152315/crashlytics/app/android:com.simprints.id/issues/9f70fa04d60773d9576c0a1c4381e56a?utm_campaign=extensions&utm_medium=SLACK&utm_source=NEW_ISSUE)  Crashlytics OOM error, revealing that OkHttp had created 300 threads. After reviewing the OkHttp [documentation](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html), I discovered that SID was incorrectly creating a new `OkHttpClient` instance for each API call instead of reusing a single instance.
### Notable changes

- Reuses `OkHttpClient` for identical auth tokens, reducing memory usage.
- Ensures thread-safe access with a lock.
- Creates new instances only when the auth token changes.
- Extracts timeouts as constants.

### Testing guidance
As the change is in the network layer then all general api calls should be checked
1. **Login & Logout** – Validate authentication/ non-authnticated API calls.  
2. **Sync (Scheduled & Manual)** – Ensure correct up/down sync behavior.  
3. **License or Firmware files Download** – Verify required files download properly.  
4. **Image Upload** – Check successful capture, upload, and retries.
### Additional work checklist

### DIFF
--- a/infra/network/src/main/java/com/simprints/infra/network/httpclient/BuildOkHttpClientUseCase.kt
+++ b/infra/network/src/main/java/com/simprints/infra/network/httpclient/BuildOkHttpClientUseCase.kt
@@ -19,8 +19,10 @@ import okio.buffer
 import java.io.IOException
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
+import javax.inject.Singleton
 
-internal class DefaultOkHttpClientBuilder @Inject constructor(
+@Singleton
+internal class BuildOkHttpClientUseCase @Inject constructor(
     @ApplicationContext private val ctx: Context,
     private val networkCache: Cache,
     private val persistentLogger: PersistentLogger,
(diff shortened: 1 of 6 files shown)


luhmirin-s (approved):
This change feels unsafe, but I cannot pin-point why :D

infra/network/src/main/java/com/simprints/infra/network/httpclient/BuildOkHttpClientUseCase.kt
```diff
@@ -40,20 +44,49 @@ internal class DefaultOkHttpClientBuilder @Inject constructor(
         const val FORCE_VERSION_HEADER = "X-Force-Version"
     }
 
-    fun get(
+    private var okHttpClient: OkHttpClient? = null
+    private var currentAuthToken: String? = null
```
BurningAXE, line 48: Just to double check - once logged in, we always use just one auth token for all calls going through OkHttp!?

meladRaouf: Yes, We use the same firebase auth token in every authenticated api request
if the token changed then the `BuildOkHttpClientUseCase` will create another OK http client

PR #1139, opened 2025-03-12
[MS-936] Fix the db locked issue by Refactor EventDatabaseFactory to create a single instance of the event DB

meladRaouf:
[MS-936 ](https://simprints.atlassian.net/browse/MS-936)

First known affected version: **it is happing in all versions 

As seen in the bug ticket and [Crashlytics](https://console.firebase.google.com/project/simprints-152315/crashlytics/app/android:com.simprints.id/issues/1e397ff40fcf20f006fc93041dc683da?utm_campaign=extensions&utm_medium=JIRA&utm_source=CRASHLYTICS_CONSOLE_MANUAL_ISSUE), we are encountering multiple errors where the events' database is locked. Upon reviewing the code, I found that we are not adhering to Room's [documentation guidelines](https://developer.android.com/training/data-storage/room#database). Specifically, As SID is creating multiple instances of the `EventRoomDatabase` object, whereas the documentation states that the database instance should be shared. Using the Android Profiler, I observed that multiple database objects are being created during the biometric capture workflow and when opening the app dashboard.  

### Notable changes

- Change EventDatabaseFactory to use `get()` instead of `build()` to retrieve the `EventRoomDatabase` instance, ensuring that the database is only built once.
- Replaced deleteDatabase() and recreateDatabaseKey() with single method `recreateDatabase()`
- Update the `EventLocalDataSource` to use `get()` for database retrieval.
- Improve `recreateDatabase()` functionality in `EventDatabaseFactory`
- Update the test suit accordingly.

### Testing guidance

* Every part of the app, including sync and biometric capture, is affected. A quick test is recommended.

### Additional work checklist

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/event/local/EventLocalDataSource.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/event/local/EventLocalDataSource.kt
@@ -22,17 +22,19 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import javax.inject.Inject
+import javax.inject.Singleton
 import kotlin.coroutines.CoroutineContext
 
+@Singleton
 internal open class EventLocalDataSource @Inject constructor(
     private val eventDatabaseFactory: EventDatabaseFactory,
     private val jsonHelper: JsonHelper,
     @DispatcherIO private val readingDispatcher: CoroutineDispatcher,
     @NonCancellableIO private val writingContext: CoroutineContext,
 ) {
-    private var eventDao: EventRoomDao = eventDatabaseFactory.build().eventDao
+    private var eventDao: EventRoomDao = eventDatabaseFactory.get().eventDao
 
-    private var scopeDao: SessionScopeRoomDao = eventDatabaseFactory.build().scopeDao
+    private var scopeDao: SessionScopeRoomDao = eventDatabaseFactory.get().scopeDao
 
     private val mutex = Mutex()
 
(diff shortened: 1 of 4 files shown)


infra/events/src/test/java/com/simprints/infra/events/event/local/EventDatabaseFactoryTest.kt
```diff
@@ -31,58 +31,68 @@ internal class EventDatabaseFactoryTest {
...
     }
 
     @Test
-    fun `test build db creates key if not exist`() = runTest {
+    fun `test get the same db instance`() = runTest {
+        // Given
+        coEvery { securityManager.getLocalDbKeyOrThrow(dbName) } returns localDbKey
+        mockkObject(EventRoomDatabase)
+        val db: EventRoomDatabase = mockk()
+        every { EventRoomDatabase.getDatabase(context, any(), dbName) } returns db
```
BurningAXE, line 51: Shouldn't this return different mocks to match actual behaviour and ensure we are not calling getDatabase() multiple times?

infra/events/src/main/java/com/simprints/infra/events/event/local/EventLocalDataSource.kt
```diff
@@ -82,16 +84,10 @@ internal open class EventLocalDataSource @Inject constructor(
         ex.let { it as? SQLiteException }?.message?.contains("file is not a database") == true
 
     private suspend fun rebuildDatabase(ex: Throwable) = mutex.withLock {
```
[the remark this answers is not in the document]
meladRaouf: done

infra/events/src/test/java/com/simprints/infra/events/event/local/EventDatabaseFactoryTest.kt
meladRaouf: I renamed the test case as it was ambiguous. The new name is  `get should return the same db instance on multiple calls` and it now verifies that the db is created only once

PR #1151, opened 2025-03-25
MS-930 Implement stateful face matcher that reduces amount of allocations

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-930)

This is an experimental implementation of the suggestion in the JIRA ticket, it could be polished to be production-ready if required.

### Notable changes

* Changing the face matches to pre-allocate native sample template memory and reuse it for each candidate. In theory, it should reduce memory and GC pressure.
* Still using a separate allocation per coroutine to avoid cross-coroutine memory sharing and potential concurrency issues. There is still a significant reduction in allocation calls.

Surprisingly the change is way less noticeable in my very basic benchmarking (running identification on 50k candidates using ROCv3 on a Lenovo Tab 8): 

| Before change  | After change |
|--------|--------|
| 114780ms | 92866ms  |
| 113032ms | 110177ms |
| 107323ms | 110511ms |
| 110483ms | 95675ms  |
| 111383ms | 104624ms |
| 113065ms | 109125ms |
| 112612ms | 110679ms |
| Avg: **111811** | Avg: **104808**

As part of this task, I also investigated the feasibility of the same refactoring for fingerprints and decided against it for the following reasons:

* SimAfis matching supports 1:N on the library level and takes the candidate list as a parameter, therefore it is highly likely that the same allocation optimisation is happening under the hood (or at least it should be done on the library level).
* NEC wraps templates with a custom data class and there is no way to control allocations. Therefore even if we cache the template list, it would not prevent unnecessary allocations.
* Fingerprint BioSDK setup is significantly more complex and would require a lot of changes for no/negligible benefit.

### Additional work checklist


face/infra/bio-sdk-resolver/src/main/java/com/simprints/face/infra/biosdkresolver/FaceBioSDK.kt
```diff
@@ -3,10 +3,15 @@ package com.simprints.face.infra.biosdkresolver
...
 interface FaceBioSDK {
     val initializer: FaceBioSdkInitializer
     val detector: FaceDetector
-    val matcher: FaceMatcher
+
     val version: String
+    val templateFormat: String
+    val matcherName: String
+
+    fun createMatcher(samples: List<FaceSample>): FaceMatcher
```
BurningAXE, line 16: Can we name these `probeSamples` as it's unclear what samples these are until you look at the implementation?

face/infra/bio-sdk-resolver/src/test/java/com/simprints/face/infra/biosdkresolver/RocV1BioSdkTest.kt
```diff
@@ -0,0 +1,22 @@
...
+
+
+import com.google.common.truth.Truth.*
+import io.mockk.*
+import org.junit.Test
+
+class RocV1BioSdkTest {
+
+
+    private lateinit var resolveFaceBioSdkUseCase: RocV1BioSdk
```
BurningAXE, line 11: This name looks wrong - it's not the resolver but he bio SDK itself!?

face/infra/bio-sdk-resolver/src/test/java/com/simprints/face/infra/biosdkresolver/RocV3BioSdkTest.kt
```diff
@@ -0,0 +1,21 @@
+package com.simprints.face.infra.biosdkresolver
+
+
+import com.google.common.truth.Truth.*
+import io.mockk.*
+import org.junit.Test
+
+class RocV3BioSdkTest {
+
+    private lateinit var resolveFaceBioSdkUseCase: RocV3BioSdk
```
BurningAXE, line 10: Here, too.

face/infra/roc-v3/src/main/java/com/simprints/face/infra/rocv3/matching/RocV3Matcher.kt
```diff
@@ -1,46 +1,53 @@
...
+        val matchTemplate = roc.new_uint8_t_array(rocConstants.ROC_FACE_FAST_FV_SIZE.toInt())
+        roc.memmove(roc.roc_cast(matchTemplate), candidateTemplate)
 
         val similarity = roc.roc_embedded_compare_templates(
             probeTemplate,
-            ROC_FACE_FAST_FV_SIZE,
+            rocConstants.ROC_FACE_FAST_FV_SIZE,
             matchTemplate,
-            ROC_FACE_FAST_FV_SIZE,
+            rocConstants.ROC_FACE_FAST_FV_SIZE,
```
BurningAXE, line 43: Can we not add rocConstants 😇

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -74,7 +75,7 @@ internal class FaceMatcherUseCase @Inject constructor(
                             loadedCandidates++
                             trySend(MatcherState.CandidateLoaded)
                         }
-                        match(batchCandidates, samples)
+                        bioSdk.createMatcher(samples).use { match(it, batchCandidates) }
```
BurningAXE, line 78: This creates a new matcher for each batch. I suggest moving it outside.

BurningAXE (changes requested):

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
luhmirin-s: What do you mean by "outside"?

face/infra/roc-v3/src/main/java/com/simprints/face/infra/rocv3/matching/RocV3Matcher.kt
luhmirin-s: It is just a matter of imports since those are defined in the SDK, I can make more explicit import.

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
BurningAXE: Sorry, meant to move it outside the batch processing - i.e. above `createRanges()`. This way we'll have only one instance of the matcher that's reused by all batches.

BurningAXE (approved):

PR #1154, opened 2025-03-27
[MS-961] Remove templateQualityScore from FingerprintSample 

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-961)
Based on the [Slack discussion](https://simprints.slack.com/archives/G01B5GC5AH3/p1742801326612559), the `templateQualityScore` field is unnecessary and does not need to be stored in the records database.

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/subject/biometricref/fingerprint/ApiFingerprintTemplate.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/subject/biometricref/fingerprint/ApiFingerprintTemplate.kt
@@ -5,7 +5,6 @@ import com.simprints.core.domain.fingerprint.IFingerIdentifier
 
 @Keep
 internal data class ApiFingerprintTemplate(
-    val quality: Int,
     val template: String,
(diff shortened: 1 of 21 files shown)


luhmirin-s (approved):

infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/remote/models/subject/biometricref/fingerprint/ApiFingerprintTemplate.kt
```diff
@@ -5,7 +5,6 @@ import com.simprints.core.domain.fingerprint.IFingerIdentifier
 
 @Keep
 internal data class ApiFingerprintTemplate(
-    val quality: Int,
```
BurningAXE, line 8: Don't we need to send the template quality to BFSID?

meladRaouf: we send it as part of the `FingerprintCaptureEvent` payload 
it is sent  as part of the session in json format 
quality  is still exist in ApiFingerprintCapturePayload

PR #1161, opened 2025-04-10
[MS-951]: Feature/add room module

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-951)

### Notable changes

* renamed EnrolmentRecordLocalDataSource to RealmEnrolmentRecordLocalDataSource
* intrduced RoomEnrolmentRecordLocalDataSource and added a dummy SelectEnrolmentRecordLocalDataSourceUseCase for now to control which db to be  active

### Testing guidance

* Test all DB operations Add, delete and update.
* All DB retrieval scenarios count subjects and list face or fingerprint templates on verify or identify flows  

### Additional work checklist


infra/core/src/main/java/com/simprints/core/domain/face/FaceSample.kt
```diff
@@ -19,9 +19,7 @@ data class FaceSample(
 
         other as FaceSample
 
-        if (!template.contentEquals(other.template)) return false
```
BurningAXE, line 22: Curious why it was implemented like this before!?

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbBiometricTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+        Index(value = [SUBJECT_ID_COLUMN]),
+        Index(value = [UUID_COLUMN], unique = true),
+    ],
+)
+@Suppress("ArrayInDataClass")
+data class DbBiometricTemplate(
+    @PrimaryKey(autoGenerate = true) val id: Long = 0,
+    val uuid: String = "",
+    val subjectId: String = "",
+    val fingerIdentifier: Int? = null, // for fingerprint samples only
```
luhmirin-s, line 32: I suggest calling this column just `val identifier: String` to keep it a bit more abstract, as we could be using it for "left/right" ear in the future.

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbSubject.kt
```diff
@@ -0,0 +1,72 @@
...
+         * - `attendantId` is second to support filtering by project AND attendant.
+         * - `createdAt` is last for efficient `ORDER BY createdAt` after filtering.
+         * Covers queries like: WHERE projectId = ? AND attendantId = ? ORDER BY createdAt
+         */
+        Index(value = [DbSubject.PROJECT_ID_COLUMN, DbSubject.ATTENDANT_ID_COLUMN, DbSubject.CREATED_AT_COLUMN]),
+
+        /**
+         * : Index solely on Project ID.
+         * Useful for frequent queries filtering *only* by projectId.
+         * Covers queries like: WHERE projectId = ?
```
luhmirin-s, line 47: Is that even a problem since there should be data from a single project on the device at one time, IIRC?

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/SubjectDao.kt
```diff
@@ -0,0 +1,46 @@
...
+    suspend fun deleteBiometricSample(uuid: String)
+
+    @RawQuery
+    suspend fun deleteSubjects(query: SupportSQLiteQuery): Int
+
+    @Query("SELECT * FROM DbSubject WHERE subjectId = :subjectId")
+    suspend fun getSubject(subjectId: String): SubjectBiometrics?
+
+    @RawQuery
+    suspend fun loadSubjects(query: SupportSQLiteQuery): List<SubjectBiometrics>
```
luhmirin-s, line 39: It is a bit concerning that query building is disjointed from the returning model definition; sooner or later, a bug will be caused by changing one place and forgetting another.

I am not sure what would be the best way to fix it, tho.

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrollmentRecordLocalDataSourceTest.kt
```diff
@@ -0,0 +1,1511 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.util.Date
+
+@RunWith(RobolectricTestRunner::class)
+class RoomEnrollmentRecordLocalDataSourceTest {
+    companion object {
+        // Constants remain the same
```
luhmirin-s, line 34: By definition :D

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrollmentRecordLocalDataSourceTest.kt
```diff
@@ -0,0 +1,1511 @@
...
+        assertThat(loaded).hasSize(1)
+        val createdSubject = loaded[0]
+        assertThat(createdSubject.subjectId).isEqualTo(subject3P1WithBoth.subjectId)
+        assertThat(createdSubject.faceSamples).hasSize(1)
+        assertThat(createdSubject.faceSamples).containsExactly(faceSample2)
+        assertThat(createdSubject.fingerprintSamples).hasSize(1)
+        assertThat(createdSubject.fingerprintSamples).containsExactly(fingerprintSample2)
+    }
+
+    @Test(expected = IllegalArgumentException::class) // Reverted to JUnit exception check
```
luhmirin-s, line 281: I remember there were some issues which resulted in incorrect test failures where exceptions were actually expected. Have you checked if test fails without this declaration?

luhmirin-s (commented):
Amazing job! I must admit that I just skimmed on THE test file :D

I have couple of small to mid-size concerns in the comments, the general approach LGTM.

build-logic/convention/src/main/kotlin/common/BuildTypeConfigurations.kt
```diff
@@ -37,7 +37,13 @@ fun Project.configureDbEncryptionBuild() {
             }
 
             getByName(BuildTypes.DEBUG) {
-                buildConfigField("Boolean", "DB_ENCRYPTION", "$propDbEncrypted")
+                // Set false only if running tests
```
BurningAXE, line 40: Hm, why did you change debug builds to be encrypted?

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrollmentRecordLocalDataSourceTest.kt
meladRaouf: :D

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RealmFaceSampleConverter.kt
```diff
@@ -1,16 +1,16 @@
 package com.simprints.infra.enrolment.records.repository.local.models
 
 import com.simprints.core.domain.face.FaceSample
-import com.simprints.infra.enrolment.records.realm.store.models.DbFaceSample
+import com.simprints.infra.enrolment.records.realm.store.models.DbFaceSample as RealmFaceSample
 
-internal fun DbFaceSample.fromDbToDomain(): FaceSample = FaceSample(
+internal fun RealmFaceSample.toDomain(): FaceSample = FaceSample(
```
BurningAXE, line 6: Didn't we agree to get rid of the incorrect 'sample' terminology? Or is it planned for a later stage?

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/SubjectBiometrics.kt
```diff
@@ -0,0 +1,14 @@
...
+import androidx.room.Relation
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.SUBJECT_ID_COLUMN
+
+data class SubjectBiometrics(
+    @Embedded val subject: DbSubject,
+    @Relation(
+        parentColumn = SUBJECT_ID_COLUMN,
+        entityColumn = SUBJECT_ID_COLUMN,
+    )
+    val biometricSamples: List<DbBiometricTemplate>,
```
BurningAXE, line 13: biometricTemplates? 😇

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RoomSubjectConverter.kt
```diff
@@ -0,0 +1,19 @@
...
+
+internal fun SubjectBiometrics.toDomain() = Subject(
+    subjectId = subject.subjectId.toString(),
+    projectId = subject.projectId,
+    attendantId = subject.attendantId.asTokenizableEncrypted(),
+    moduleId = subject.moduleId.asTokenizableEncrypted(),
+    createdAt = subject.createdAt?.toDate(),
+    updatedAt = subject.updatedAt?.toDate(),
+    fingerprintSamples = biometricSamples.filterNot { it.fingerIdentifier == null }.map { it.toFingerprintSample() },
+    faceSamples = biometricSamples.filter { it.fingerIdentifier == null }.map { it.toFaceSample() },
```
BurningAXE, line 16: As we add more modalities this will become an unreliable way to distinguish templates. We'll need something like a type enum. Maybe it's best if we introduce it from the start and avoid changes later.
And once we have it, maybe we won't need to have separate fields for each modality as we'll be able to quickly filter the needed templates. I'm much less certain about this second point, though.

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
```diff
@@ -0,0 +1,226 @@
...
+              "createdAt"
+            ],
+            "orders": [],
+            "createSql": "CREATE INDEX IF NOT EXISTS `index_DbSubject_projectId_attendantId_createdAt` ON `${TABLE_NAME}` (`projectId`, `attendantId`, `createdAt`)"
+          },
+          {
+            "name": "index_DbSubject_projectId",
+            "unique": false,
+            "columnNames": [
+              "projectId"
```
BurningAXE, line 79: If yes, don't we need it here, too?

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
```diff
@@ -0,0 +1,226 @@
...
+            "unique": false,
+            "columnNames": [
+              "subjectId",
+              "format"
+            ],
+            "orders": [],
+            "createSql": "CREATE INDEX IF NOT EXISTS `index_DbBiometricTemplate_subjectId_format` ON `${TABLE_NAME}` (`subjectId`, `format`)"
+          },
+          {
+            "name": "index_DbBiometricTemplate_subjectId",
```
BurningAXE, line 150: If you already have an index that has the same column(s) as prefix, you don't need a separate one. You can filter on `index_DbBiometricTemplate_subjectId_format` by subjectid just as efficiently: https://sqlite.org/queryplanner.html#_multi_column_indices

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
```diff
@@ -0,0 +1,226 @@
...
+        ],
+        "primaryKey": {
+          "autoGenerate": true,
+          "columnNames": [
+            "id"
+          ]
+        },
+        "indices": [
+          {
+            "name": "index_DbBiometricTemplate_subjectId_format",
```
BurningAXE, line 140: Do we need this index if we have a separate lookup table?

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/FingerIdentifierAndTemplate.kt
```diff
@@ -0,0 +1,9 @@
+package com.simprints.infra.enrolment.records.room.store.models
+
+import androidx.room.ColumnInfo
+
+@Suppress("ArrayInDataClass")
+data class FingerIdentifierAndTemplate(
```
BurningAXE, line 6: That's one weird name and class! Maybe I'm missing something as I don't get it's purpose.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+        val sqlQuery = buildSubjectQuery(query)
+        subjectDao.loadSubjects(SimpleSQLiteQuery(sqlQuery.first, sqlQuery.second.toTypedArray())).map { it.toDomain() }
+    }
+
+    override suspend fun count(
+        query: SubjectQuery,
+        dataSource: BiometricDataSource,
+    ): Int = withContext(dispatcherIO) {
+        val (sql, queryArgs) = buildCountQuery(query)
+        subjectDao.countSubjects(SimpleSQLiteQuery(sql, queryArgs.toTypedArray()))
```
BurningAXE, line 61: Why don't we return the query from the buildQuery functions in the needed final format?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/EnrolmentRecordsStoreModule.kt
```diff
@@ -54,6 +50,9 @@ class IdentityDataSourceModule {
         context = context,
         dispatcher = dispatcher,
     )
+
+    @Provides
+    fun provideBatchSize(): Int = BATCH_SIZE
```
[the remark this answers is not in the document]
meladRaouf: Done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+            sqlBuilder.append(" WHERE ").append(whereClauses.joinToString(" AND "))
+        }
+        val resultQuery = sqlBuilder.toString()
+        return Pair(resultQuery, args)
+    }
+
+    private fun buildBiometricTemplatesQuery(
+        query: SubjectQuery,
+        range: IntRange,
+    ): Pair<String, List<Any?>> {
```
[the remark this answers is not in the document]
meladRaouf: It is recommended to pass the args separately to let room cash and reuse queries

infra/enrolment-records/repository/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
     id("simprints.infra")
     id("kotlin-parcelize")
+    id("simprints.library.room")
```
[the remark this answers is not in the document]
meladRaouf: I only need the RoomDatabase.withTransaction from` implementation(libs.androidX.room.core)`
will remove the plugin declaration and use the normal dependency for room

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbBiometricTemplate.kt
meladRaouf: done

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbSubject.kt
meladRaouf: Sorry I didn't get your point

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrollmentRecordLocalDataSourceTest.kt
meladRaouf: Yes I tested it now and it crashes when I removed the expected = IllegalArgumentException::class

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RealmFaceSampleConverter.kt
meladRaouf: I see from the git history that DB_ENCRYPTION was always true in debug builds

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RoomBiometricTemplateConverter.kt
```diff
@@ -0,0 +1,38 @@
+package com.simprints.infra.enrolment.records.repository.local.models
+
+import com.simprints.core.domain.face.FaceSample
+import com.simprints.core.domain.fingerprint.FingerprintSample
+import com.simprints.core.domain.fingerprint.IFingerIdentifier
+import com.simprints.infra.enrolment.records.room.store.models.DbBiometricTemplate
+
+internal fun DbBiometricTemplate.toFingerprintSample(): FingerprintSample = FingerprintSample(
+    id = uuid,
+    fingerIdentifier = IFingerIdentifier.entries[fingerIdentifier!!], // fingerIdentifier shouldn't be null for fingerprint samples
```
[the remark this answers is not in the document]
meladRaouf: I have added the FACE_MODALITY and FINGERPRINT_MODALITY

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/SubjectBiometrics.kt
meladRaouf: Done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RoomSubjectConverter.kt
meladRaouf: I have added an extra string field "modality"

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
```diff
@@ -0,0 +1,226 @@
...
+          ]
+        },
+        "indices": [
+          {
+            "name": "index_DbSubject_projectId_moduleId_createdAt",
+            "unique": false,
+            "columnNames": [
+              "projectId",
+              "moduleId",
+              "createdAt"
```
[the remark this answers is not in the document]
meladRaouf: Yes, for pagination we have to sort to avoid inconsistent results.

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
meladRaouf: done

build-logic/convention/src/main/kotlin/common/BuildTypeConfigurations.kt
meladRaouf: I see from the git history that DB_ENCRYPTION was always true in debug builds

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/FingerIdentifierAndTemplate.kt
meladRaouf: It is unused and deleted

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/SubjectsDatabaseFactory.kt
```diff
@@ -0,0 +1,57 @@
...
+        )
+    } catch (t: Throwable) {
+        Simber.e("Error creating subject database", t)
+        throw t
+    }
+
+    private fun getOrCreateKey() = try {
+        secureLocalDbKeyProvider.getLocalDbKeyOrThrow(DB_NAME)
+    } catch (t: Throwable) {
+        t.message?.let { Simber.d(it) }
```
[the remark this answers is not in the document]
meladRaouf: the first use of `secureLocalDbKeyProvider.getLocalDbKeyOrThrow` will always  throw `MissingLocalDatabaseKeyException`

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+
+    // Private Helper Methods
+    private suspend fun <T> loadBiometricIdentities(
+        query: SubjectQuery,
+        range: IntRange,
+        getSampleFormat: (SubjectQuery) -> String?,
+        createIdentity: (subjectId: String, samples: List<DbBiometricTemplate>) -> T,
+        onCandidateLoaded: () -> Unit,
+    ): List<T> = withContext(dispatcherIO) {
+        val sampleFormat = getSampleFormat(query)
```
BurningAXE, line 162: Can't we just pass the sampleFormat instead of a function that gets it :)

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+        subjectDao
+            .loadSamples(SimpleSQLiteQuery(sqlQuery.first, sqlQuery.second.toTypedArray()))
+            .map { (subjectId, templates) ->
+                onCandidateLoaded()
+                createIdentity(subjectId, templates)
+            }
+    }
+
+    private suspend fun createSubject(
+        subject: SubjectDomain,
```
BurningAXE, line 179: Since the DB model is DbSubject, for consistency the domain model should be DomainSubject

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+        if (dbSubject != null) {
+            // delete face and finger samples
+            val referencesToDelete = action.referenceIdsToRemove.toSet()
+            require(
+                referencesToDelete.size != dbSubject.biometricTemplates.size ||
+                    action.faceSamplesToAdd.isNotEmpty() ||
+                    action.fingerprintSamplesToAdd.isNotEmpty(),
+            ) {
+                val errorMsg = "Cannot delete all samples for subject ${action.subjectId} without adding new ones"
+                Simber.i("[updateSubject] $errorMsg", tag = ROOM_RECORDS_DB)
```
BurningAXE, line 238: This should probably be Simber.e() as this is not supposed to happen. At the same time, if it happens, will lead to discrepancy between BFSID and SID data.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+        }
+    }
+
+    private suspend fun deleteSubject(subjectId: String) {
+        Simber.d("[deleteSubject] Deleting subject: $subjectId", tag = ROOM_RECORDS_DB)
+        subjectDao.deleteSubject(subjectId)
+    }
+
+    // SQL Query Builder Methods
+    private fun buildSubjectQuery(query: SubjectQuery): Pair<String, List<Any?>> {
```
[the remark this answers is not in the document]
BurningAXE: Agree. All these manual strings will be prone to errors (some of which may go under the radar as less often used) once we start changing the DB schemas.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,400 @@
...
+                subjectDao.insertBiometricSamples(templatesToAdd)
+            }
+        } else {
+            Simber.i("[updateSubject] Subject ${action.subjectId} not found for update", tag = ROOM_RECORDS_DB)
+        }
+    }
+
+    private suspend fun deleteSubject(subjectId: String) {
+        Simber.d("[deleteSubject] Deleting subject: $subjectId", tag = ROOM_RECORDS_DB)
+        subjectDao.deleteSubject(subjectId)
```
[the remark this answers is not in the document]
meladRaouf: I am using `onDelete = ForeignKey.CASCADE`
The CASCADE option ensures that when the parent row is deleted, all child rows referencing it will also be automatically deleted.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: Yes, you are right, this should go to the catalytics non-fatal section. 
Done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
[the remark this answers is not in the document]
BurningAXE: Right, already forgot what I saw on Thursday :D

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: the code 
```
 require(
                referencesToDelete.size != dbSubject.biometricTemplates.size ||
                    action.faceSamplesToAdd.isNotEmpty() ||
                    action.fingerprintSamplesToAdd.isNotEmpty(),
            )
```
will report a crash in crashlytics so no need to change to Simber.e()

infra/enrolment-records/repository/build.gradle.kts
```diff
@@ -11,9 +11,11 @@ dependencies {
     implementation(project(":infra:config-store"))
     implementation(project(":infra:auth-store"))
     implementation(project(":infra:enrolment-records:realm-store"))
+    implementation(project(":infra:enrolment-records:room-store"))
     implementation(project(":infra:events"))
 
     implementation(libs.libsimprints)
     implementation(libs.retrofit.core)
     implementation(libs.jackson.core)
+    implementation(libs.androidX.room.core)
```
luhmirin-s, line 20: Why is this necessary? Shouldn't all direct DB interaction be in room-store module?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/EnrolmentRecordRepositoryImpl.kt
```diff
@@ -10,52 +10,37 @@ import com.simprints.infra.enrolment.records.realm.store.exceptions.RealmUniniti
...
-
+    private val selectEnrolmentRecordLocalDataSource: SelectEnrolmentRecordLocalDataSourceUseCase,
+    @DispatcherIO private val dispatcher: CoroutineDispatcher,
+    @EnrolmentBatchSize private val batchSize: Int,
+) : EnrolmentRecordRepository {
     private val prefs = context.getSharedPreferences(PREF_FILE_NAME, Context.MODE_PRIVATE)
 
     companion object {
-        private const val BATCH_SIZE = 80
+        const val BATCH_SIZE = 80
```
luhmirin-s, line 37: Since the batch size is an external argument, it does not make sense to store the constant in this class anymore.

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbSubject.kt
luhmirin-s: Query by only the project ID is same as query without a filter, since all of the records on the device should be from the same project. Or is there a case when there are records from multiple projects?

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/Modality.kt
```diff
@@ -0,0 +1,6 @@
+package com.simprints.infra.enrolment.records.room.store.models
+
+enum class Modality {
```
luhmirin-s, line 3: Please add a huge warning not to change the order or remove any lines here since the DB data depends on the ordinal values.

Or use text fields to avoid those issues.

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
```diff
@@ -0,0 +1,233 @@
...
+          },
+          {
+            "fieldPath": "subjectId",
+            "columnName": "subjectId",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "identifier",
+            "columnName": "identifier",
```
luhmirin-s, line 111: Any reason not to use a string field instead for better clarity?

infra/enrolment-records/repository/build.gradle.kts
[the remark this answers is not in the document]
luhmirin-s: Separate dependency is worse since the whole point of using convention plugins is to keep major dependencies encapsulated. The direct database action should be in the respective module instead, IMO.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/models/RoomSubjectConverter.kt
luhmirin-s: String or int?

build-logic/convention/src/main/kotlin/common/BuildTypeConfigurations.kt
luhmirin-s: It was likely that way to catch encryption issues early instead of chasing them in production :)

meladRaouf: Unfortunately I couldn't get the SQL Cipher .so files to be loaded in the unit test target

infra/enrolment-records/repository/build.gradle.kts
meladRaouf: If you check the RoomEnrolmentRecordLocalDataSource, you'll see that I use database.withTransaction from the Room library to ensure all actions, including the deletion, are executed atomically.

luhmirin-s: Shouldn't that be part of the data source instead?

luhmirin-s: I.e. the part of the data-source module.

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/DbSubject.kt
meladRaouf: So far, this has been the case, but if you check the standalone SID, you'll notice that different projects could be stored in the same database since there's no login mechanism. It is safer to have the project id index

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
meladRaouf: you are right for the table `DbBiometricTemplate`
I don't need 
`Index(value = [SUBJECT_ID_COLUMN]),` 
As I already have amd  still need 
'Index(value = [SUBJECT_ID_COLUMN, FORMAT_COLUMN]),`

meladRaouf: done

infra/enrolment-records/room-store/schemas/com.simprints.infra.enrolment.records.room.store.SubjectsDatabase/1.json
meladRaouf: Yes as per the `Explain query plan` this is needed 
you can think of loading the subjects in the identification process as two queries, the first one gets the list of subjects and uses the lokup table 
the second one gets the list of templates and uses the BiometricTemplate table and needs this index

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: yes, this would be simpler.
Done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/EnrolmentRecordRepositoryImpl.kt
meladRaouf: done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
```diff
@@ -0,0 +1,163 @@
...
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.MODULE_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.PROJECT_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.SUBJECT_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.SUBJECT_TABLE_NAME
+import com.simprints.infra.enrolment.records.room.store.models.DbSubjectTemplateFormatMap.Companion.FORMAT_MAP_TABLE_NAME
+import jakarta.inject.Inject
+
+internal class RoomEnrolmentRecordQueryBuilder @Inject constructor() {
+    companion object {
+        const val TABLE_SUBJECT = SUBJECT_TABLE_NAME
```
luhmirin-s, line 18: This feels a bit redundant. Are those constants redefined locally just for the tests?

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilderTest.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.junit.Test
+
+class RoomEnrolmentRecordQueryBuilderTest {
+    private val builder = RoomEnrolmentRecordQueryBuilder()
+
+    @Test
+    fun `buildSubjectQuery with empty query returns base query`() {
+        val query = SubjectQuery()
+        val result = builder.buildSubjectQuery(query)
+        assertThat(result.sql).contains("SELECT * FROM ${RoomEnrolmentRecordQueryBuilder.TABLE_SUBJECT} s")
```
luhmirin-s, line 14: This might be counterintuitive, but I suggest avoiding constants in the test cases and using plain strings instead. 

There could be a case when a developer makes a typo in the constant definition, and it would be propagated to the test cases and make them green. In this case, the plain strings could act as a double-check against such minor typos.

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/Modality.kt
```diff
@@ -0,0 +1,9 @@
+package com.simprints.infra.enrolment.records.room.store.models
+
+// **IMPORTANT**: Do NOT change the order of this enum as it is used in the database by index.
```
luhmirin-s, line 3: Just realised that we could also do
```
enum class Modality(val id: Int) { 
   FINGERPRINT(0), FACE(1) 

   companion object {
     private val lookupTable = values.associateBy { it.id } 
 
     fun forId(id: Int): Modality? = lookupTable.getOrNull { id == it.id }
   }
}
```

And use the explicit ID instead of ordinal to make it safer.

luhmirin-s: A similar approach could be made for the identifiers field (with a dedicated enum class) to ensure correctness and not rely on not changing order in any of the enums in the future.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
meladRaouf: Yes, the intention was to avoid passing the internal DB constants to the tests, but I've removed them now.

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilderTest.kt
meladRaouf: Done

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/Modality.kt
meladRaouf: This is a brilliant idea. It allows us to safely continue using integers instead of strings. 
Done

luhmirin-s: Then the warning is not required, at least for the modality table.

infra/core/src/main/java/com/simprints/core/domain/fingerprint/IFingerIdentifier.kt
```diff
@@ -5,15 +5,26 @@ import com.simprints.core.ExcludedFromGeneratedTestCoverageReports
...
-    LEFT_THUMB,
-    LEFT_INDEX_FINGER,
-    LEFT_3RD_FINGER,
-    LEFT_4TH_FINGER,
-    LEFT_5TH_FINGER,
+// **IMPORTANT**: Do NOT change the order of this enum as it is used in the database by index.
+// Changing the order of the entries in this enum will lead to data corruption or mismatches
+// when retrieving or storing data in the database. Always append new entries at the end.
+enum class IFingerIdentifier(
+    val id: Int,
```
luhmirin-s, line 12: I would be very careful about changing the classes in the core/domain as they are likely used directly in the event and sample marshalling and a bunch of other places, adding fields to them at this point might break it. It would be safer to have a dedicated "identifier" enum in the DB module and map values from one to another based on the value and modality.

infra/enrolment-records/room-store/src/main/java/com/simprints/infra/enrolment/records/room/store/models/Modality.kt
meladRaouf: I removed the warning from modality class

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
```diff
@@ -0,0 +1,125 @@
...
+                $whereClause
+                LIMIT $pageSize
+            ) B USING(subjectId) where A.format ='$format'
+            """.trimIndent()
+        println(sql)
+        println("----")
+        return SimpleSQLiteQuery(sql, args.toTypedArray())
+    }
+
+    fun buildWhereAndOrderByClause(
```
luhmirin-s, line 75: Could those be 2 separate functions to be cleaner in purpose?

luhmirin-s (approved):

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
```diff
@@ -0,0 +1,125 @@
...
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.ATTENDANT_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.MODULE_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.PROJECT_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.SUBJECT_ID_COLUMN
+import com.simprints.infra.enrolment.records.room.store.models.DbSubject.Companion.SUBJECT_TABLE_NAME
+import jakarta.inject.Inject
+
+internal class RoomEnrolmentRecordQueryBuilder @Inject constructor() {
+    fun buildSubjectQuery(query: SubjectQuery): SimpleSQLiteQuery {
+        // require format not to be set for subject query and guid to use the buildBiometricTemplatesQuery instead
```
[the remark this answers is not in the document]
BurningAXE: I think it's a type and was meant to be "guid**e** to use the buildBiometricTemplatesQuery instead"!?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+            "SELECT COUNT(DISTINCT S.$SUBJECT_ID_COLUMN) FROM $SUBJECT_TABLE_NAME S $whereClause"
+        }
+        return SimpleSQLiteQuery(sql, args.toTypedArray())
+    }
+
+    fun buildBiometricTemplatesQuery(
+        query: SubjectQuery,
+        pageSize: Int,
+        lastSeenSubjectId: String? = null,
```
BurningAXE, line 54: I'd name this `afterSubjectId` 🙈 👀

BurningAXE: Hm, I see the query already has `afterSubjectId` and we copy this parameter there. Why the different name then?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
```diff
@@ -0,0 +1,136 @@
...
+            args.add(it)
+        }
+        query.subjectIds?.takeIf { it.isNotEmpty() }?.let {
+            clauses.add("${subjectAlias}$SUBJECT_ID_COLUMN IN (${it.joinToString(",") { "?" }})")
+            args.addAll(it)
+        }
+        query.afterSubjectId?.let {
+            clauses.add("${subjectAlias}$SUBJECT_ID_COLUMN > ?")
+            args.add(it)
+        }
```
BurningAXE, line 102: Since `subjectId`, `subjectIds` and `afterSubjectId` are mutually exclusive - do we want to have them in a when to prevent the callers from shooting themselves in the foot?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,310 @@
...
+    ): ReceiveChannel<List<T>> {
+        var lastSeenSubjectId: String? = null
+        var lastOffset = 0
+        return loadIdentitiesConcurrently(
+            ranges = ranges,
+            dispatcher = dispatcherIO,
+            parallelism = PARALLELISM,
+            scope = scope,
+        ) { range ->
+            require(lastOffset == range.first) {
```
BurningAXE, line 157: There should be a +1 from `lastOffset` because `range.last` is inclusive as is `range.first`
(This works in the current main because of the bug we discovered in #1194)

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,310 @@
...
+                )
+                val sql = "DELETE FROM DbSubject $whereClause"
+                subjectDao.deleteSubjects(SimpleSQLiteQuery(sql, args.toTypedArray()))
+            }
+        }
+    }
+
+    override suspend fun deleteAll() {
+        Simber.i("[deleteAll] Deleting all subjects.", tag = ROOM_RECORDS_DB)
+        subjectDao.deleteSubjects(SimpleSQLiteQuery("DELETE FROM DbSubject"))
```
BurningAXE, line 216: Same here.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,310 @@
...
+                subjectDao.deleteBiometricSample(it.uuid)
+            }
+            val templatesToAdd =
+                action.faceSamplesToAdd.map { it.toRoomDb(action.subjectId) } +
+                    action.fingerprintSamplesToAdd.map { it.toRoomDb(action.subjectId) }
+            if (templatesToAdd.isNotEmpty()) {
+                subjectDao.insertBiometricSamples(templatesToAdd)
+            }
+        } else {
+            Simber.i(
```
BurningAXE, line 299: Simber.e()?

infra/enrolment-records/repository/src/test/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrollmentRecordLocalDataSourceTest.kt
```diff
@@ -0,0 +1,1537 @@
...
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import java.util.Date
+import java.util.UUID
+
+@RunWith(RobolectricTestRunner::class)
+class RoomEnrollmentRecordLocalDataSourceTest {
```
BurningAXE, line 35: The class (and the file) has double L while the source class has only a single one (as is the accepted naming convention).

alex-vt (approved):

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,310 @@
...
+     * Don't use this method if the query contains format fields (faceSampleFormat or fingerprintSampleFormat).
+     * instead, use loadFaceIdentities or loadFingerprintIdentities methods.
+     */
+    override suspend fun load(query: SubjectQuery): List<DomainSubject> = withContext(dispatcherIO) {
+        if (query.hasUntokenizedFields == true) {
+            Simber.d(
+                "[load] Query has untokenized fields, returning empty list as all records are tokenized.",
+                tag = ROOM_RECORDS_DB,
+            )
+            return@withContext emptyList()
```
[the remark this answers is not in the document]
meladRaouf: If you trace this method call, you'll see that it gets invoked frequently during every remote config update.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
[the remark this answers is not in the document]
meladRaouf: yes it is guide not guid

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordQueryBuilder.kt
meladRaouf: changed it to after subject id

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: Can you merge your PR into main ?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: It seems to be a common scenario where BFSID attempts to update a subject that hasn't been down-synced yet or has been deleted because its module was deselected by the user.

BurningAXE: Isn't the first a problem?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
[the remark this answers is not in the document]
meladRaouf: This situation can occur during a configuration refresh. When the tokenization keys become available, SID attempts to tokenize any un-tokenized records. As a result, the load(query: SubjectQuery) method will be called frequently with hasUntokenizedFields == true.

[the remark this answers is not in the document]
meladRaouf: We can remove this call once Realm has been completely removed, since the load method is implemented the same way in both the Realm and Room data sources.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
meladRaouf: I updated the logging to log this case as an error with Simber.e()

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -86,20 +78,15 @@ internal class FaceMatcherUseCase @Inject constructor(
                 scope = this,
                 onCandidateLoaded = {
                     loadedCandidates.incrementAndGet()
-                    this.trySend(MatcherState.CandidateLoaded)
+                    runBlocking {
```
BurningAXE, line 81: Do we really need this to be blocking?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/commcare/CommCareIdentityDataSource.kt
```diff
@@ -159,7 +167,6 @@ internal class CommCareIdentityDataSource @Inject constructor(
             .query(caseDataUri, null, null, null, null)
             ?.use { caseDataCursor ->
                 val subjectActions = getSubjectActionsValue(caseDataCursor)
-                Simber.d(subjectActions)
```
BurningAXE, line 162: This is often useful

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RoomEnrolmentRecordLocalDataSource.kt
```diff
@@ -0,0 +1,303 @@
...
+        query: SubjectQuery,
+        dataSource: BiometricDataSource,
+    ): Int = withContext(dispatcherIO) {
+        subjectDao.countSubjects(queryBuilder.buildCountQuery(query))
+    }
+
+    /**
+     * Loads face identities in paged ranges.
+     */
+    override fun loadFaceIdentities(
```
BurningAXE, line 82: Should these be suspending, too?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/commcare/CommCareIdentityDataSource.kt
meladRaouf: Reverted back

fingerprint/infra/simprints-bio-sdk/src/main/java/com/simprints/fingerprint/infra/biosdkimpl/matching/SimAfisMatcher.kt
```diff
@@ -56,8 +56,6 @@ internal class SimAfisMatcher @Inject constructor(
     ): List<MatchResult> {
         val simAfisCandidates = candidates.map { it.toSimAfisPerson() }
 
-        println("Matching ${simAfisCandidates.size} candidates using all ${jniLibAfis.getNbCores()} cores")
```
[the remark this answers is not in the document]
meladRaouf: I don't like using the println, but I reverted it back

PR #1170, opened 2025-04-15
[MS-973] Fix sorting logic in MatchResultSet to handle equal confidence scores

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-973)

When two subjects have the same confidence score during matching, only one is returned in the identification response. Please refer to the ticket for more details.

First known affected version: **All recent versions**

* Attached two videos one for the problem and the fix

### Notable changes

*  Improve sorting logic in MatchResultSet to handle equal confidence scores by subject ID

### Testing guidance

* Use the project cGKXBnPkLGKtK0V2TUxo, which contains many duplicate ROCv3 and simatcher subjects, and verify that subjects with the same confidence score are all included in the identification results.

### Additional work checklist

https://github.com/user-attachments/assets/889549ed-dbeb-4386-b411-a9fde9dae7e6

https://github.com/user-attachments/assets/2dcc8454-676d-4d9c-af32-ac13e1a195ad

### DIFF
--- a/feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
+++ b/feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
@@ -8,10 +8,9 @@ internal class MatchResultSet<T : MatchResultItem>(
 ) {
     private var lowestConfidence: Float = 0f
 
-    private val treeSet = TreeSet { o1: T, o2: T ->
-        // Reverse order for descending sort
-        -1 * o1.confidence.compareTo(o2.confidence)
-    }
+    private val treeSet = TreeSet(
+        compareByDescending<T> { it.confidence }.thenByDescending { it.subjectId },
+    )
 
     fun add(element: T): MatchResultSet<T> {
         if (lowestConfidence > element.confidence) {
--- a/feature/matcher/src/test/java/com/simprints/matcher/usecases/MatchResultSetTest.kt
+++ b/feature/matcher/src/test/java/com/simprints/matcher/usecases/MatchResultSetTest.kt
@@ -46,4 +46,23 @@ class MatchResultSetTest {
             ),
         )
     }
+
+    // On equal confidence scores sort by id
+    @Test
+    fun `Stores results sorted descending by confidence and id`() {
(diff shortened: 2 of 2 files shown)


feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
```diff
@@ -8,10 +8,9 @@ internal class MatchResultSet<T : MatchResultItem>(
...
-    private val treeSet = TreeSet { o1: T, o2: T ->
-        // Reverse order for descending sort
-        -1 * o1.confidence.compareTo(o2.confidence)
-    }
+    private val treeSet = TreeSet(
+        compareByDescending<T> { it.confidence }.thenByDescending { it.subjectId },
+    )
 
     fun add(element: T): MatchResultSet<T> {
         if (lowestConfidence > element.confidence) {
```
BurningAXE, line 16: Hm, am I reading this wrong or if we first add a high confidence score (say 99), then we would not add any lower results and return a list of just 1 result? Is this the behaviour that we want?

BurningAXE: Ah, `lowestConfidence` is updated only when the set is full. That's confusing...

alex-vt: It looks like if `confidence` and `subjectId` are now a composite comparison key, so should work as intended.

alex-vt (approved):

feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
BurningAXE: Here we are comparing confidence only. The confusion comes from the fact that lowestConfidence is initialized with 0 and updated only when the set is full.

meladRaouf: This is a classic LeetCode problem 🥇. If we only need the top N elements, fully sorting the entire list is unnecessary and can be avoided to enhance performance.

luhmirin-s (approved):

PR #1179, opened 2025-04-24
[MS-120] Skipping non-existent privacy notice download 404 logging

alex-vt:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-120)

### Notable changes

Not user-facing, only affects logging:

When an API call for downloading a privacy notice occurs, and returns 404 (not found) - not logging this.


infra/config-store/src/main/java/com/simprints/infra/config/store/ConfigRepositoryImpl.kt
```diff
@@ -36,6 +37,7 @@ internal class ConfigRepositoryImpl @Inject constructor(
     companion object {
         @VisibleForTesting
         const val PRIVACY_NOTICE_FILE = "privacy_notice"
+        private const val NOT_FOUND_STATUS_CODE = 404
```
luhmirin-s, line 40: I would suggest using the codes defined in the HTTP library to skip on re-defining them in our code base.

luhmirin-s (approved):

PR #1187, opened 2025-05-12
feature/migrate-records-from-realm-to-room

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-991)

### Notable changes

1. **Introduced Realm to Room Migration Worker**  

2. **Controlled via Remote Config and Batched Execution**  
   The migration is gated by the following remote config flags:
   - `recordsDbMigrationFromRealmEnabled`: Enables or disables the migration logic.
   - `recordsDbMigrationFromRealmMaxRetries`: Specifies the maximum number of retry attempts (default: 5).  
   Migration only runs when down sync is not active and processes data in batches of 500 records.

3. **Dual Write Support During Migration**  
   Any data added to Realm during migration will also be written to Room to ensure consistency.

4. **Failure Handling**  
   If migration fails after the maximum retry limit, the Room database will be deleted and the migration will restart from the beginning on the next eligible run.

5. **Retry Logic and Logging**  
   The worker retries up to the configured limit (`recordsDbMigrationFromRealmMaxRetries`) on failure. All migration events and errors are logged to Google Analytics for visibility and debugging.

6. **Benchmark Results**  
   Based on internal testing, a batch size of 500 records offers optimal performance when each record includes **two biometric templates**. If records contain **more than two templates**, a smaller batch size may be required to avoid memory or performance issues. Below are the results for migrating 50,000 records:

   | Batch Size | Time Taken (ms) | Time Taken (minutes) |
   |------------|------------------|-----------------------|
   | 1000       | 364,476          | 6.07                  |
   | 100        | 281,573          | 4.69                  |
   | 200        | 253,284          | 4.22                  |
   | 300        | 248,007          | 4.13                  |
   | **500**    | **235,598**      | **3.93**              |
   | 600        | 239,528          | 4.00                  |

**The [design document](https://docs.google.com/document/d/17m9xxLZbfahRWx1G8AULy6R2aTJYv13ZYot3_JO73qs/edit?tab=t.0#heading=h.ignevd9587y) has been updated to reflect the recent changes in the migration strategy.**

### Testing guidance

*should be added soon

### Additional work checklist


infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
```diff
@@ -0,0 +1,100 @@
...
+                crashlyticsLog("migrated in ${time.inWholeMilliseconds}")
+            } else {
+                crashlyticsLog("No records to migrate. Skipping migration.")
+            }
+            realmToRoomMigrationFlagsStore.updateStatus(MigrationStatus.COMPLETED)
+            realmToRoomMigrationFlagsStore.resetRetryCount()
+            return@withContext success()
+        } catch (e: Exception) {
+            Simber.e("Migration failed: ${e.message}", e)
+            realmToRoomMigrationFlagsStore.incrementRetryCount()
```
alex-vt, line 75: If I haven't missed anything: a device may reach the `MAX_MIGRATION_RETRIES` limit here. Then the `RealmToRoomMigrationScheduler` will return at line 30 and won't be able to retry at line 36, leading to a migration lockout (device remaining stuck with the old DB). Is this the intended behavior?

The alternative is, ultimately, an unrestricted allowed number of attempts to migrate, but then we need to be careful about scheduling and exponential backoff, so we don't put the job on an endless loop.

I'd sugguest to:
* leave the max attempt capped as a safer option,
* make the `MAX_MIGRATION_RETRIES` a custom config param,
* intruduce a condition for the staged rollout of the release where this code will end up in: check the logs for the retry counts before deciding to advance to the next stage.

Then we watch the logs. If too many devices reach the max migration retries, we decide how to sort it out - by a hotfix, just more retries, a reinstall via user support, etc. Then eventually we bump the max retries in the custom configs for those impacted devices - so they can use the additional attempts to migrate.

meladRaouf: I agree that making the migration retry limit configurable via a remote config key would be beneficial. It offers full control over the behaviour in the future, especially since the specific causes of potential migration failures are not yet known. A remote configuration mechanism would allow for dynamic adjustments without requiring code changes. I will update the code and the design doc

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/workers/EventDownSyncDownloaderWorker.kt
```diff
@@ -65,6 +67,12 @@ internal class EventDownSyncDownloaderWorker @AssistedInject constructor(
     private suspend fun getDownSyncOperation() = eventDownSyncScopeRepository.refreshState(downSyncOperationInput)
 
     override suspend fun doWork(): Result = withContext(dispatcher) {
+        // Check if the migration is in progress before starting the sync
+        if (realmToRoomMigrationFlagsStore.isMigrationInProgress()) {
+            // this will make the worker retry in 5 minutes
```
alex-vt, line 72: Maybe it would be more clear if there's a reference to what the value of 5 minutes is coming from

infra/resources/src/main/res/values/strings.xml
```diff
@@ -438,6 +438,7 @@
 
     <!-- Notifications -->
     <string name="notification_sync_description">Sync in progress</string>
+    <string name="notification_migration_description">Storage optimization in progress</string>
```
alex-vt, line 441: :+1:

alex-vt (approved):

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationFlagsStore.kt
```diff
@@ -0,0 +1,78 @@
...
+        prefs.getInt(KEY_MIGRATION_RETRIES, 0) <
+            configRepo.getProjectConfiguration().experimental().recordsDbMigrationFromRealmMaxRetries
+    }
+
+    suspend fun isMigrationGloballyEnabled(): Boolean = mutex.withLock {
+        configRepo.getProjectConfiguration().experimental().recordsDbMigrationFromRealmEnabled
+    }
+
+    suspend fun isDownSyncInProgress(): Boolean = mutex.withLock {
+        prefs.getBoolean(KEY_DOWN_SYNC_STATUS, false)
```
luhmirin-s, line 69: 1. I think it would be worth checking for the sync process as a whole since the down-sync always comes after the up-sync. There could be a moment where this is triggered during a lengthy up-sync, and then down-sync could disrupt it.
2. Aren't there already existing mechanisms to track the sync status in the codebase?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
```diff
@@ -0,0 +1,105 @@
...
+        const val WORK_NAME = "room-to-realm-migration-worker-v1"
+    }
+
+    override val tag: String
+        get() = REALM_DB_MIGRATION.name
+
+    lateinit var project: Project
+
+    override suspend fun doWork(): Result = withContext(dispatcher) {
+        project = configRepo.getProject()
```
luhmirin-s, line 47: Would it be overkill to completely cancel sync workers during migration start and reschedule them afterwards?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
```diff
@@ -0,0 +1,105 @@
...
+            realmToRoomMigrationFlagsStore.updateStatus(MigrationStatus.FAILED)
+            roomDataSource.deleteAll()
+            return@withContext fail(e, message = "Migration failed: ${e.message}")
+        }
+    }
+
+    override val notificationDescription = IDR.string.notification_migration_description
+
+    private suspend fun processRecords() {
+        val scope = CoroutineScope(dispatcher + SupervisorJob())
```
luhmirin-s, line 88: Wouldn't this make the record processing uncooperative if the worker has been cancelled for some reason? Since the scope is not being explicitly cancelled anywhere and it is not a direct child of the workers scope.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RealmEnrolmentRecordLocalDataSource.kt
```diff
@@ -257,4 +261,36 @@ internal class RealmEnrolmentRecordLocalDataSource @Inject constructor(
         }
         return realmQuery
     }
+
+    suspend fun loadAllSubjectsInBatches(
+        batchSize: Int,
+        scope: CoroutineScope,
+    ): ReceiveChannel<List<Subject>> {
```
luhmirin-s, line 268: Can this be done with flows instead?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
luhmirin-s: Am I right in assuming that the version part in the shared pref keys is to be able to reset the migration counter after a hotfix?

meladRaouf: Yes, I added the version to leave room for changes in future releases, if needed.

infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/down/workers/EventDownSyncDownloaderWorker.kt
meladRaouf: I have updated the comment to be 
As per DEFAULT_BACKOFF_INTERVAL_MINUTES the worker will retry in 5 minutes

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationFlagsStore.kt
meladRaouf: 1. I want to block down sync during the records DB migration because it's the only operation that can modify the records DB, I wanted to prevent any updates while the migration is in progress to avoid data corruption or inconsistent state. Upsync, on the other hand, operates solely on the events DB and doesn’t interact with the records DB, so it can safely continue. Additionally, if there’s an issue with the migration or the new DB implementation, keeping upsync functional allows users to safely upload all sessions. This ensures data isn’t lost and makes it easier to recover by simply instructing users to log out or reinstall the app.
2. WorkManager has a method to query worker status, but using it would create a circular dependency on the `event-sync` module as I need the tag ID or worker class name. To avoid that, I chose a simpler solution: storing the status in shared preferences.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
meladRaouf: I don’t want to interfere with  upsyncs to ensure users can still upload their sessions without interruption.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RealmEnrolmentRecordLocalDataSource.kt
meladRaouf: Yes flow would be better and it won't need a scope param

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
luhmirin-s: IMO, it would be simpler to just skip a sync cycle than worry about partial up/down syncs.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationFlagsStore.kt
luhmirin-s: But both up and down syncs are linked, so blocking one also blocks the whole sync anyway.

infra/enrolment-records/repository/build.gradle.kts
```diff
@@ -17,4 +17,5 @@ dependencies {
     implementation(libs.libsimprints)
     implementation(libs.retrofit.core)
     implementation(libs.jackson.core)
+    implementation(libs.androidX.room.core)
```
luhmirin-s, line 20: I guess this is temporary until the rebase onto #1161, right?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/InsertRecordsDuringMigrationUseCase.kt
```diff
@@ -0,0 +1,28 @@
+package com.simprints.infra.enrolment.records.repository.local.migration
+
+import com.simprints.infra.config.store.models.Project
+import com.simprints.infra.enrolment.records.repository.domain.models.SubjectAction
+import com.simprints.infra.enrolment.records.repository.local.RoomEnrolmentRecordLocalDataSource
+import com.simprints.infra.logging.LoggingConstants.CrashReportTag.REALM_DB_MIGRATION
+import com.simprints.infra.logging.Simber
+import jakarta.inject.Inject
```
luhmirin-s, line 8: Weird import.

luhmirin-s (approved):

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
meladRaouf: removed

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationFlagsStore.kt
meladRaouf: The good thing is that the downsync worker is last in the chain, so retrying it won't affect the execution of the preceding workers.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/InsertRecordsDuringMigrationUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+// Should be completely removed when all users are migrated to Room
+internal class InsertRecordsDuringMigrationUseCase @Inject constructor(
+    private val realmToRoomMigrationFlagsStore: RealmToRoomMigrationFlagsStore,
+    private val roomEnrolmentRecordLocalDataSource: RoomEnrolmentRecordLocalDataSource,
+) {
+    suspend operator fun invoke(
+        subjectAction: SubjectAction.Creation,
+        project: Project,
+    ) {
+        // if the realm to room migration is in progress, we need to insert the records in the new db
```
BurningAXE, line 19: What if the migration fails?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationScheduler.kt
```diff
@@ -0,0 +1,82 @@
...
+        when (currentStatus) {
+            MigrationStatus.FAILED -> {
+                log("Migration status is FAILED and can retry. Scheduling worker.")
+                enqueueMigrationWork()
+            }
+            MigrationStatus.NOT_STARTED,
+            MigrationStatus.IN_PROGRESS,
+            -> { // If app was killed during these states
+                log("Migration status is $currentStatus. Scheduling worker to start.")
+                enqueueMigrationWork()
```
BurningAXE, line 42: All 3 states do the same except for the logs but we've already reported the current status. So, is there really a need for this whole `when`?

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/InsertRecordsDuringMigrationUseCase.kt
BurningAXE: Ah, I get it now - the new records will go to Realm and we also want to insert them in Room, too.
Then I guess the comment can be made clearer:
"// if the realm to room migration is in progress, we need to insert the records in the new db**, too**"

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/EnrolmentRecordRepositoryImpl.kt
```diff
@@ -137,19 +137,24 @@ internal class EnrolmentRecordRepositoryImpl @Inject constructor(
...
 
-    override suspend fun deleteAll() = localDataSource.deleteAll()
+    override suspend fun deleteAll() = selectEnrolmentRecordLocalDataSource().deleteAll()
 
     override suspend fun performActions(
         actions: List<SubjectAction>,
         project: Project,
-    ) = localDataSource.performActions(actions, project)
+    ) {
+        actions.filterIsInstance<SubjectAction.Creation>().forEach {
```
BurningAXE, line 155: This is problematic - what happens with the updates and deletes? They will modify records in the old DB but we may have already migrated those records. I understand that this is not expected as downsync is blocked and we don't currently support local updates or deletes. But at the very least I'd put an error log here if that's attempted so we know something got wrong.

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationScheduler.kt
```diff
@@ -0,0 +1,82 @@
...
+            else -> {
+                // No action needed for COMPLETED or any other state
+            }
+        }
+    }
+
+    private fun enqueueMigrationWork() {
+        val constraints = Constraints
+            .Builder()
+            .setRequiredNetworkType(NetworkType.NOT_REQUIRED)
```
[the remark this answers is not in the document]
meladRaouf: I removed it

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/RealmEnrolmentRecordLocalDataSource.kt
```diff
@@ -300,4 +302,32 @@ internal class RealmEnrolmentRecordLocalDataSource @Inject constructor(
         }
         return realmQuery
     }
+
+    fun loadAllSubjectsInBatches(batchSize: Int): Flow<List<Subject>> = channelFlow {
+        var lastId: RealmUUID? = null
```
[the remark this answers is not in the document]
meladRaouf: done

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/InsertRecordsDuringMigrationUseCase.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import com.simprints.infra.config.store.models.Project
+import com.simprints.infra.enrolment.records.repository.domain.models.SubjectAction
+import com.simprints.infra.enrolment.records.repository.local.RoomEnrolmentRecordLocalDataSource
+import com.simprints.infra.logging.LoggingConstants.CrashReportTag.REALM_DB_MIGRATION
+import com.simprints.infra.logging.Simber
+import javax.inject.Inject
+
+// Should be completely removed when all users are migrated to Room
+internal class InsertRecordsDuringMigrationUseCase @Inject constructor(
```
BurningAXE, line 11: How about `InsertRecords**InRoom**DuringMigrationUseCase` to make it clearer as I was confused initially?

meladRaouf: Yes, this is a better name

infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/EnrolmentRecordRepositoryImpl.kt
```diff
@@ -135,19 +140,38 @@ internal class EnrolmentRecordRepositoryImpl @Inject constructor(
...
-    ) = localDataSource.performActions(actions, project)
+    ) {
+        logUpdatesDuringMigration(actions)
+        actions.filterIsInstance<SubjectAction.Creation>().forEach {
+            insertRecordsInRoomDuringMigration(it, project)
+        }
+        selectEnrolmentRecordLocalDataSource().performActions(actions, project)
+    }
+
+    private suspend fun logUpdatesDuringMigration(actions: List<SubjectAction>) {
```
BurningAXE, line 165: IMO this is better placed inside `InsertRecordsInRoomDuringMigrationUseCase` so logic is more encapsulated and both can be deleted at once. But that's a personal preference.

meladRaouf: Yes this would be cleaner

PR #1169, opened 2025-04-15
[MS-949] Introduce Concurrency for Face and Fingerprint Matching

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-949)
Will be released in: **2025.2.0**

### Notable changes

- Implement producer-consumer pattern using coroutines and channels to process candidate batches concurrently.
- Configure the number of consumer coroutines based on the available processors to maximize efficiency.
- Refactor FaceMatcherUseCase and FingerprintMatcherUseCase to utilize the new concurrent matching approach.
- Update tests to accommodate the changes in the matching logic.

### Testing guidance
Evaluate verification and identification using face and fingerprint modalities with both small and large candidate sets for matching.
### Additional work checklist

* [x] Effect on other features and security has been considered
* [ ] Design document marked as "In development" (if applicable)
* [ ] External (Gitbook) and internal (Confluence) Documentation is up to date (or ticket created)
* [ ] Test cases in Testiny are up to date (or ticket created)
* [ ] Other teams notified about the changes (if applicable)

PR #1194, opened 2025-05-20
[MS-949] Improve concurrency in matching

BurningAXE:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-949)

### Notable changes

This PR builds upon [1169](https://github.com/Simprints/Android-Simprints-ID/pull/1169) by making the following changes:
- Ranges are created in such a way so that their count is multiples of the CPU cores. This way all ranges can be processed in parallel at all times, minimizing CPU idling and maximizing efficiency. Fore example for a CPU with 4 cores, totalCount will be divided into 4, 8, 12, etc. ranges depending on the size of totalCount. A maximum size of 2000 is imposed for range size.
- MatchResultSet was refactored to handle concurrency as the previous implementation lead to crashes if modified concurrently
- The loadedCandidates counter was also modified to be thread-safe as previously it had inconsistent behaviour
- flows in the face/fingerprint matcher use cases are started on `Dispatchers.IO` instead of the caller's `Dispatchers.Main`

### Testing guidance

I've tested this extensively with CoSync but would like someone to double check it with the internal DB. Trying out different totalCount's would also be nice to ensure no edge cases are left out.

### Additional work checklist

### DIFF
--- a/feature/matcher/src/main/java/com/simprints/matcher/usecases/CreateRangesUseCase.kt
+++ b/feature/matcher/src/main/java/com/simprints/matcher/usecases/CreateRangesUseCase.kt
@@ -1,45 +1,51 @@
 package com.simprints.matcher.usecases
 
 import javax.inject.Inject
+import kotlin.math.ceil
 
 internal class CreateRangesUseCase @Inject constructor() {
     /**
      * Creates a list of ranges to be used for batch processing.
-     * Range size is increased dynamically to ensure that first couple of batches are small
-     * to speed up initial reads, then it increases to ensure that the last batches are not too small.
-     *
-     * For example with minBatchSize = 10, returned batches will be 10, 10, 20, 30, 40, 50, 50 in size (if the total allows).
+     * The number of ranges will be a multiple of the available processors to ensure
+     * efficient parallel processing.
+     * Range sizes are adjusted to not exceed MAX_BATCH_SIZE.
      */
     operator fun invoke(
         totalCount: Int,
-        minBatchSize: Int = DEFAULT_BATCH_SIZE,
+        availableProcessors: Int = Runtime.getRuntime().availableProcessors()
     ): List<IntRange> {
+        if (totalCount <= 0) return emptyList()
+
+        // Calculate how many multiples of processors we need to keep batches under MAX_BATCH_SIZE
+        val batchesPerProcessor = ceil(totalCount.toDouble() / (availableProcessors * MAX_BATCH_SIZE)).toInt().coerceAtLeast(1)
+        val totalBatches = availableProcessors * batchesPerProcessor
+
+        // Calculate the base batch size and remainder for even distribution
+        val baseBatchSize = totalCount / totalBatches
+        val remainder = totalCount % totalBatches
+
         val ranges = mutableListOf<IntRange>()
-        var index = 1
-
-        var nextBatchSize = minBatchSize
         var start = 0
-        var end = nextBatchSize
 
-        while (start < totalCount) {
-            if (end > totalCount) {
-                end = totalCount
-            }
-            ranges.add(start..end)
+        // Create ranges with sizes distributed as evenly as possible
+        for (i in 0 until totalBatches) {
+            // Add 1 to batch size for the first 'remainder' batches to distribute the remainder evenly
+            val batchSize = baseBatchSize + if (i < remainder) 1 else 0
+            val end = (start + batchSize).coerceAtMost(totalCount)
+
+            ranges.add(start until end)
             start = end
-            end += nextBatchSize
 
-            // Make sure next batch is increased
-            nextBatchSize = minBatchSize + (minBatchSize * index.coerceIn(1, 4))
-            index++
+            if (start >= totalCount) break
         }
+
         return ranges
     }
 
     companion object {
         /**
-         * Experimentally determined batch size that works well for most cases.
+         * Maximum size for a batch to avoid huge memory consumption
          */
-        private const val DEFAULT_BATCH_SIZE = 1000
+        private const val MAX_BATCH_SIZE = 2000
     }
 }
--- a/feature/matcher/src/main/java/com/simprints/matcher/usecases/CreateRangesUseCase.kt
+++ b/feature/matcher/src/main/java/com/simprints/matcher/usecases/CreateRangesUseCase.kt
@@ -1,9 +1,12 @@
 package com.simprints.matcher.usecases
 
+import com.simprints.core.AvailableProcessors
 import javax.inject.Inject
 import kotlin.math.ceil
 
-internal class CreateRangesUseCase @Inject constructor() {
+internal class CreateRangesUseCase @Inject constructor(
+    @AvailableProcessors private val availableProcessors: Int,
+) {
     /**
      * Creates a list of ranges to be used for batch processing.
      * The number of ranges will be a multiple of the available processors to ensure
@@ -12,7 +15,6 @@ internal class CreateRangesUseCase @Inject constructor() {
      */
     operator fun invoke(
         totalCount: Int,
-        availableProcessors: Int = Runtime.getRuntime().availableProcessors()
     ): List<IntRange> {
         if (totalCount <= 0) return emptyList()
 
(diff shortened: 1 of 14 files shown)


feature/matcher/src/test/java/com/simprints/matcher/usecases/CreateRangesUseCaseTest.kt
```diff
@@ -10,42 +10,259 @@ class CreateRangesUseCaseTest {
...
+        // When
+        val result = useCase(50000, 4)
+
+        // Then
+        // Need ceiling(50000/(4*2000)) = 7 batches per processor = 28 total batches
+        // Base size = 50000/28 = 1785, remainder = 20
+        // First 20 batches get size 1786, remaining 8 batches get size 1785
+        val expected = mutableListOf<IntRange>()
+        var start = 0
+        for (i in 0 until 28) {
```
luhmirin-s, line 240: Do you have tests for this computation?

feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
```diff
@@ -1,40 +1,41 @@
...
-        if (treeSet.size > maxSize) {
-            treeSet.pollLast()
+        skipListSet.add(element)
+        if (skipListSet.size > maxSize) {
+            skipListSet.pollLast()
 
             // Not that the set is full, we can skip adding elements
             // with confidence lower than the current lowest
-            lowestConfidence = treeSet.last().confidence
+            lowestConfidence.set(skipListSet.last().confidence)
```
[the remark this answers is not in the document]
luhmirin-s: Copilot has the right idea, but it is off. 

If we want to make this data structure thread safe, the whole `add()` function should be blocking/atomic; otherwise, multiple threads could start racing to edit the lowest confidence between `get()` and `set()`. 

Another option would be to change how the set is used and prevent the concurrent addition completely. At the moment within `consumeAndMatch()`, one set is populated sequentially, and then it is `addAll()`-ed to a final set, which can be called from multiple coroutines at once. Instead, we could make `consumeAndMatch()` return the intermediate resulting set of that particular call and then merge all of them sequentially. 
Compared to the matching step, this would likely be negligible computation and memory cost (each set is only up to 10 elements), but would provide the best concurrency safety (as there is no concurrency) without explicit locking and atomic integers.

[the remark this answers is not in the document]
BurningAXE: Good call! I prefer to make `add()` thread-safe instead of relying on the caller doing things in a particular way.

feature/matcher/src/main/java/com/simprints/matcher/usecases/CreateRangesUseCase.kt
```diff
@@ -1,45 +1,51 @@
...
-                end = totalCount
-            }
-            ranges.add(start..end)
+        // Create ranges with sizes distributed as evenly as possible
+        for (i in 0 until totalBatches) {
+            // Add 1 to batch size for the first 'remainder' batches to distribute the remainder evenly
+            val batchSize = baseBatchSize + if (i < remainder) 1 else 0
+            val end = (start + batchSize).coerceAtMost(totalCount)
+
+            ranges.add(start until end)
```
meladRaouf, line 36: This should be `ranges.add(start..end)` instead of `ranges.add(start until end)` to maintain consistency with the previous implementation and prevent skipping one record in each range.

luhmirin-s: Isn't `..` inclusive on both ends, and therefore there will actually be an overlap if it is used?

feature/matcher/src/test/java/com/simprints/matcher/usecases/CreateRangesUseCaseTest.kt
```diff
@@ -10,42 +10,259 @@ class CreateRangesUseCaseTest {
...
+    @Test
+    fun `should handle 2500 candidates with 4 processors`() {
+        // When
+        val result = useCase(2500, 4)
+
+        // Then
+        // 4 processors, batches under 2000 each, so 4 total batches
+        // Base size = 2500/4 = 625, remainder = 0
+        assertThat(result).containsExactly(
+            0 until 625,
```
luhmirin-s, line 137: Test cases before this one use the `rangeTo` operator; it would be great to have it consistent, does not matter which way.

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -59,12 +57,18 @@ internal class FaceMatcherUseCase @Inject constructor(
...
+
         Simber.i("Matching candidates", tag = crashReportTag)
         send(MatcherState.LoadingStarted(expectedCandidates))
-        val ranges = createRanges(expectedCandidates)
+
+        // When using local DB loadedCandidates = expectedCandidates
+        // However, when using CommCare as data source, loadedCandidates < expectedCandidates
+        // as it's count function does not take into account filtering criteria
+        val loadedCandidates = AtomicInteger(0)
+        val availableProcessors = Runtime.getRuntime().availableProcessors()
```
luhmirin-s, line 68: Processor count should be injected/provided from the outside.

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
```diff
@@ -88,8 +92,8 @@ internal class FaceMatcherUseCase @Inject constructor(
         }
         // Wait for all to complete
         consumerJobs.forEach { it.join() }
-        send(MatcherState.Success(resultSet.toList(), loadedCandidates, bioSdk.matcherName))
-    }
+        send(MatcherState.Success(resultSet.toList(), loadedCandidates.get(), bioSdk.matcherName))
+    }.flowOn(dispatcherIO)
```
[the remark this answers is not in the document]
BurningAXE: > All read operations already run on the IO dispatcher using withContext(dispatcherIO)
That's the problem - they don't! Reading was happening on the main thread because it used the scope passed from the ViewModel! This is why the UI wasn't updating counts until all reading was done.

feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
```diff
@@ -1,40 +1,48 @@
...
         compareByDescending<T> { it.confidence }.thenByDescending { it.subjectId },
     )
 
     fun add(element: T): MatchResultSet<T> {
-        if (lowestConfidence > element.confidence) {
-            // skip adding if the last element is greater than the current element
+        // Use a lock to ensure thread safety during the entire add operation
+        lock.withLock {
+            // Only perform this optimization when we know the set is at max capacity
+            if (skipListSet.size >= maxSize && lowestConfidence.get() > element.confidence) {
```
luhmirin-s, line 23: Are atomic reference for "lowestConfidence" and concurrent collection required if the whole block is synchronised?

luhmirin-s (approved):

meladRaouf (approved):

feature/matcher/src/main/java/com/simprints/matcher/usecases/FaceMatcherUseCase.kt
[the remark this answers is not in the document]
meladRaouf: I think the flow should run on a background dispatcher, with only the database read operations running on the I/O dispatcher.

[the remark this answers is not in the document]
BurningAXE: That would be the official recommendation. However, in my testing IO provided better performance. I have no explanation why but I don't see any harm, either 🤷

feature/matcher/src/main/java/com/simprints/matcher/usecases/MatchResultSet.kt
BurningAXE: I'm tempted to dismiss this concern for structures with max 10 results. However, if you feel strongly about this, we can do a quick benchmark and verify TreeSet indeed works fine in real concurrent situations (CoSync reading with N threads)!?

alexandr-simprints (approved):

PR #1203, opened 2025-05-29
MS-989 SimFace in dev builds only

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-989)

### Notable changes

* Only bundle the SimFace in the debug build. Other build types have stub implementation to make the DI happy. 
* Opted not to bother with splitting configuration into build types as it would be extremely complex for almost no return - the `simFace` field in the config should only be provided to dev projects by the backend, so there is little to no risk of it affecting the release build. The code itself does not add any noticeable difference in the APK file.
* Also, bump SimFace dependency to a version that does not bundle test images.

### APK breakdown

Both APKs were generated from the same code/commit:

Screenshot 2025-05-29 at 11 24 13

### Testing guidance

* Create dev, staging and release builds from the same commit and check the APK sizes and contents using the analyze tool.

### Additional work checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -69,7 +69,7 @@ roc_wrapper_version = "1.23.0"
 roc_wrapper-v3_version = "3.1.0"
 nec_version = "1.5.0"
 secugen_version = "1.1.0"
-simface = "2025.2.0"
+simface = "2025.2.1"
 
 junit_version = "4.13.2"
 junit_ext_version = "1.2.1"


face/infra/bio-sdk-resolver/src/main/java/com/simprints/face/infra/biosdkresolver/ResolveFaceBioSdkUseCase.kt
```diff
@@ -1,28 +1,16 @@
...
-                .getProjectConfiguration()
-                .face
-                ?.rankOne
-                ?.version
-                ?.takeIf { it.isNotBlank() } // Ensures version is not null or empty
-            requireNotNull(version) { "FaceBioSDK version is null or empty" }
-            if (version == rocV3BioSdk.version) rocV3BioSdk else rocV1BioSdk
-        }
+        FaceConfiguration.BioSdk.SIM_FACE -> resolveSimFaceVersionUseCase()
+        FaceConfiguration.BioSdk.RANK_ONE -> resolveRankOneVersionUseCase()
```
BurningAXE, line 14: It's a matter of preference but do we really need to extract something as simple as the version resolving? It seems the overhead is higher than the code itself.

luhmirin-s: It is not necessary, but it has the benefit of isolating each SDK's logic. 

Honestly, I was just trying a lot of things to minimise the amount of classes that we would need to duplicate per build type and this is a by-product of the experimentation. We can simplify it when the need for build types is gone.

meladRaouf (approved):

PR #1220, opened 2025-06-11
[MS-1024] Add validation for Realm to Room migration consistency

meladRaouf:
When the SQLite database files are deleted and the app attempts to access the database, the SQLite library automatically recreates empty files. Although deleting the database files during migration is not an expected scenario, I’ve added a validation step to fail the migration if the record counts do not match.


infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/local/migration/RealmToRoomMigrationWorker.kt
```diff
@@ -102,7 +102,19 @@ internal class RealmToRoomMigrationWorker @AssistedInject constructor(
...
+        validateRealmToRoomMigration()
         // log room db info
         crashlyticsLog("[RealmToRoomMigrationWorker] ${roomDataSource.getLocalDBInfo()}")
     }
+
+    private suspend fun validateRealmToRoomMigration() {
+        val realmCount = realmDataSource.count(SubjectQuery())
+        val roomCount = roomDataSource.count(SubjectQuery())
+        if (realmCount != roomCount) {
+            throw IllegalStateException(
```
BurningAXE, line 114: This would fail the whole migration and reattempt it, right?

meladRaouf: Yes, it would increment the retry count and re-enqueue the worker with an appropriate backoff time.

BurningAXE (approved):

PR #1221, opened 2025-06-12
MS-122 Secure prefs everywhere

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-122)

### Notable changes

* Moving a couple of remaining cases of shared prefs usage to the encrypted counterpart to avoid accidentally leaking some IDs/domain. The risk was always very low, but this ticket has been in the backlog for too long by now.
* Did not move the language helper since it is fine as-is (it stores only the language code), and the performance penalty from encryption might affect app startup times. 

### Testing guidance

* Use the app as usual - upsync progress should be reported correctly.

### Additional work checklist

### DIFF
--- a/infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProviderImpl.kt
+++ b/infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProviderImpl.kt
@@ -6,17 +6,19 @@ import androidx.annotation.VisibleForTesting
 import androidx.core.content.edit
 import com.simprints.infra.logging.Simber
 import com.simprints.infra.network.BuildConfig
+import com.simprints.infra.security.SecurityManager
 import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 internal class BaseUrlProviderImpl @Inject constructor(
-    @ApplicationContext context: Context,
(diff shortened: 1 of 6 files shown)


alex-vt (approved):

infra/network/src/main/java/com/simprints/infra/network/url/BaseUrlProviderImpl.kt
```diff
@@ -26,13 +28,28 @@ internal class BaseUrlProviderImpl @Inject constructor(
...
+    @Deprecated("Data has been migrated to secure prefs. Delete after there are no users below 2025.3.0")
+    private val prefs: SharedPreferences = context.getSharedPreferences(LEGACY_PREF_FILE_NAME, Context.MODE_PRIVATE)
+    private val securePrefs = securityManager.buildEncryptedSharedPreferences(SECURE_PREF_FILE_NAME)
 
-    override fun getApiBaseUrl(): String = prefs
+    /**
+     * Ensures that data has been migrated to secure prefs before accessing it.
+     */
+    private fun getSecurePrefs(): SharedPreferences {
+        if (prefs.contains(API_BASE_URL_KEY)) {
```
BurningAXE, line 39: Instead of checking the old prefs (and always reading 2 files as a consequence) would it make sense to first check the securePrefs for the flag and read old prefs only if it's missing?

PR #1254, opened 2025-07-09
MS-941/MS-1058 Update sync configuration to the new structure

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-941)

### Notable changes

* CoSync changes introduced additional changes to project configuration, and since the sample upload config also relies on the same version, baseline config changes for both features have to be implemented more or less together. This PR contains only the changes in the existing configuration structure and does not add any co-sync specific fields:
  * `synchronization.down` properties moved to dedicated `synchronization.down.simprints` object. All usages are redirected to Simprints' specific config. For now the `down.simprints` is considered non-optional since it is used across modules. @BurningAXE I leave it to you to make it optional when adding co-sync configuration.
  * `synchronization.frequency` moved to `synchronization.down.simprints.frequency` and `synchronization.up.simprints.frequency`. All usages are defaulted to the "down.simprints" frequency, unless it explicitly requires "up". 
  * Due to the frequencies being in multiple places, the enum has been moved to the upper level in all layers (domain, proto and API)
* Did not add any co-sync configuration fields. @BurningAXE, that part is for you to implement. 
* This PR is based on sample upload prep work and therefore assumes that the samples config changes have been merged already (relevant for property indexes in proto files)

### Testing guidance

* Run the app, attempt configuration sync. 
* Install an older version, sync the config, then update to this PR and sync the config again.

### Additional work checklist

### DIFF
--- a/infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiSynchronizationConfiguration.kt
+++ b/infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiSynchronizationConfiguration.kt
@@ -4,35 +4,34 @@ import androidx.annotation.Keep
 import com.simprints.core.domain.tokenization.asTokenizableEncrypted
 import com.simprints.infra.config.store.models.DownSynchronizationConfiguration
 import com.simprints.infra.config.store.models.DownSynchronizationConfiguration.Companion.DEFAULT_DOWN_SYNC_MAX_AGE
+import com.simprints.infra.config.store.models.Frequency
 import com.simprints.infra.config.store.models.SampleSynchronizationConfiguration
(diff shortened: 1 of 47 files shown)


infra/config-store/src/main/java/com/simprints/infra/config/store/remote/models/ApiSynchronizationConfiguration.kt
```diff
@@ -99,16 +100,27 @@ internal data class ApiSynchronizationConfiguration(
 
     @Keep
     data class ApiDownSynchronizationConfiguration(
+        val simprints: ApiSimprintsDownSynchronizationConfiguration,
```
ybourgery, line 103: This is an optional field now, have you not done it because you will let Milen doing it?

PR #1255, opened 2025-07-10
MS-1058 Implementation of signed url sample upload

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1058)

### Notable changes

* Changed the image repository interface to accommodate for the extra information required for sample upload with signed URLs. 
* Moved sample file path creation into the module to ensure consistent paths and have an option to break that path back into usable metadata about the sample (session ID, modality and sample ID).
* Refactored the existing Firestore image uploader out of the ImageRepositoryImpl to allow for different approaches to reading files from the disk:
  * Firestore - reads and uploads files one by one in sequence until all are uploaded.
  * Signed URL - reads files in batches to calculate size and MD5 values, requests the upload URLs for the whole batch with a request, then proceeds to upload files one by one within the batch until all batches are uploaded.
* **Important** - encrypted file input stream that we are getting from the local disc can only be read once and cannot be reset, therefore any BODY-level network logging (both chucker and okhttp) will consume the stream to print it out, and the actual file upload will fail.

### Testing guidance

* Enable sample/image upload in Vulcan. 
* Do several biometric flows with face or fingerprint tracking sessions IDs. Wait a couple of hours until the state in sync info screen shows that images were uploaded. Check Firestore for the uploaded images by project and session IDs.
* Enable signed URL upload with custom configuration.
* Do several biometric flows with face or fingerprint tracking sessions IDs. Wait a couple of hours until the state in sync info screen shows that images were uploaded. Check the new sample storage for the uploaded images by project and session IDs.

### Additional work checklist

### DIFF
--- a/infra/images/src/main/java/com/simprints/infra/images/remote/firestore/FirestoreSampleUploader.kt
+++ b/infra/images/src/main/java/com/simprints/infra/images/remote/firestore/FirestoreSampleUploader.kt
@@ -9,6 +9,7 @@ import com.simprints.infra.images.local.ImageLocalDataSource
 import com.simprints.infra.images.metadata.ImageMetadataStore
 import com.simprints.infra.images.model.SecuredImageRef
 import com.simprints.infra.images.remote.SampleUploader
+import com.simprints.infra.logging.LoggingConstants.CrashReportTag.SAMPLE_UPLOAD
 import com.simprints.infra.logging.LoggingConstants.CrashReportTag.SYNC
 import com.simprints.infra.logging.Simber
 import kotlinx.coroutines.tasks.await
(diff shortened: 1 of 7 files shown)


infra/images/src/main/java/com/simprints/infra/images/remote/signedurl/SignedUrlSampleUploader.kt
```diff
@@ -0,0 +1,99 @@
...
+                try {
+                    prepareImageUploadData(imageRef).also {
+                        if (it == null) {
+                            allImagesUploaded = false
+                            Simber.i("Failed to read image file without exception", tag = SYNC)
+                        }
+                    }
+                } catch (t: Throwable) {
+                    allImagesUploaded = false
+                    Simber.e("Failed to read image file", t, tag = SYNC)
```
alexandr-simprints, line 58: Should we have any information from the `imageRef` here? While sensitive, we can pre-process the data so that the sensitive parts aren't exposed. This way we will be able reconstruct the error should there be any

infra/images/src/main/java/com/simprints/infra/images/remote/firestore/FirestoreSampleUploader.kt
```diff
@@ -0,0 +1,84 @@
...
+    private val configManager: ConfigManager,
+    private val authStore: AuthStore,
+    private val localDataSource: ImageLocalDataSource,
+    private val metadataStore: ImageMetadataStore,
+) : SampleUploader {
+    override suspend fun uploadAllSamples(projectId: String): Boolean {
+        val firebaseApp = authStore.getLegacyAppFallback()
+        val firebaseProjectName = firebaseApp.options.projectId
+
+        if (!firebaseProjectName.isNullOrBlank()) {
```
alexandr-simprints, line 28: Rather a personal preference: instead of negative-condition, I'd prefer to see
```
if (firebaseProjectName.isNullOrBlank()) {
  Simber.i("Firebase projectId is null")
  return false
}

// [ Continuation ]
```
Don't think it's worth refactoring now, but just something I'll be happy to see in future

infra/images/src/main/java/com/simprints/infra/images/remote/firestore/FirestoreSampleUploader.kt
```diff
@@ -0,0 +1,84 @@
...
+                            localDataSource.deleteImage(imageRef)
+                            metadataStore.deleteMetadata(imageRef.relativePath)
+                        } else {
+                            allImagesUploaded = false
+                            Simber.i("Failed to upload image without exception", tag = SYNC)
+                        }
+                    }
+                } catch (t: Throwable) {
+                    allImagesUploaded = false
+                    Simber.e("Failed to upload images", t, tag = SYNC)
```
alexandr-simprints, line 51: Same thing about more info on image:
> Should we have any information from the `imageRef` here? While sensitive, we can pre-process the data so that the sensitive parts aren't exposed. This way we will be able reconstruct the error should there be any

infra/images/src/main/java/com/simprints/infra/images/usecase/CalculateFileMd5AndSizeUseCase.kt
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        val digest = md5.digest()
+
+        CalculationResult(
+            md5 = encodingUtils.byteArrayToBase64(digest),
+            size = size,
+        )
+    } catch (e: IOException) {
+        e.printStackTrace()
```
alexandr-simprints, line 41: `Simber.e` instead?

infra/network/src/main/java/com/simprints/infra/network/httpclient/BuildOkHttpClientUseCase.kt
```diff
@@ -88,8 +88,13 @@ internal class BuildOkHttpClientUseCase @Inject constructor(
             if (!currentAuthToken.isNullOrBlank()) {
                 addInterceptor(buildAuthenticationInterceptor(currentAuthToken!!))
             }
-        }.addNetworkInterceptor(ChuckerInterceptor.Builder(ctx).build())
-        .addInterceptor(buildDeviceIdInterceptor(deviceId))
+        }.addNetworkInterceptor(
+            ChuckerInterceptor
+                .Builder(ctx)
+                // Skip sample storage domain to avoid interfering with encrypted file streaming
+                .skipDomains("storage.googleapis.com")
```
alexandr-simprints, line 95: Probably better to move to constant for a better readability in future. Out of context `storage.googleapis.com` doesn't say much. Maybe also worth adding a `[MS-1058]` reference to a comment

luhmirin-s: That would be a bit excessive, IMO. The value is not going to be reused anywhere, and moving it to a constant does not provide any benefit. The context is provided by the comment, which I can rephrase and expand to make it clearer. It is not related to the ticket directly, just a technical nuance of the logging library and file streaming.

infra/images/src/main/java/com/simprints/infra/images/remote/signedurl/SignedUrlSampleUploader.kt
luhmirin-s: What kind of information do you mean?

luhmirin-s:
@alexandr-simprints I have added more general logging around the upload to help us restore the timeline in crash logs (if it will be necessary). There isn't much else we could add to those logs, tho.

PR #1263, opened 2025-07-14
[MS-855] generate bulk of biometric records

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-855)

### Notable changes
This PR is the first part of generating the bulk data it adds a debug-only module that enables developers and testers to generate bulk biometric enrolment records directly into the local database. 

To trigger bulk data generation via an Intent, refer to the usage instructions in the module's README.

### Testing guidance

* generate biometric records and use them in verification and identification flows 

### Additional work checklist


.github/workflows/pr-checks.yml
```diff
@@ -65,6 +65,7 @@ jobs:
                 feature:alert
                 feature:exit-form
                 feature:select-subject-age-group
+                feature:data-generator
```
luhmirin-s, line 68: Would it make sense to put it under a different top-level package to separate dev/testing modules from the actual product features? Something like `testing:data-generator` or something like that.

feature/data-generator/src/main/java/com/simprints/feature/datagenerator/DataGeneratorViewModel.kt
```diff
@@ -0,0 +1,147 @@
...
+        private const val EXTRA_ENROL_LAST_COUNT = "EXTRA_ENROL_LAST_COUNT"
+        private const val EXTRA_VERIFY_COUNT = "EXTRA_VERIFY_COUNT"
+    }
+
+    private val _statusMessage = MutableLiveData<String>()
+    val statusMessage: LiveData<String> = _statusMessage
+
+    suspend fun handleIntent(intent: Intent?) {
+        Simber.i("Handling intent: ${intent?.action}", tag = TAG)
+        if (!BuildConfig.DEBUG_MODE) {
```
luhmirin-s, line 50: I think this is redundant, since it should be handled on the Gradle level.
If you want to have a redundant check just in case, I would suggest having it in the activity's onCreate so that the whole flow is stopped as soon as possible.

luhmirin-s (approved):

PR #1274, opened 2025-07-21
MS-1070 Persistent step cache fix

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1070)

First known affected version: **2025.1.0**

* The issue is with the persistent orchestration cache - the change in settings/permissions triggers full recreation of OrchestrationActivity from the persistent store. And when unmarshalling step params (that are kept as Bundles), the custom parcelable parameters in those bundles lose the type information and are restored as generic maps. 

### Notable changes

* Replace step parameters bundle with custom serialisable definition.
* Introduces separate marker interfaces for step params and results to enable serialisation to work. 
* Once the cache issue was fixed, another related issue surfaced - when a step's fragment returns result after configuration change (nav graph recreation), the result handlers were not triggered.

### Testing guidance

* Run all orchestration flows in all possible combinations.

### Additional work checklist


feature/alert/src/main/java/com/simprints/feature/alert/AlertConfigurationBuilder.kt
```diff
@@ -40,23 +40,22 @@ data class AlertConfigurationBuilder(
...
+    titleRes = this.titleRes,
+    image = this.image,
+    message = this.message,
+    messageRes = this.messageRes,
+    messageIcon = this.messageIcon,
+    leftButton = this.leftButton,
+    rightButton = this.rightButton,
+    eventType = this.eventType,
+    appErrorReason = this.appErrorReason,
 ).toBundle()
```
BurningAXE, line 59: Should this still be converted to Bundle?

feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorViewModel.kt
```diff
@@ -209,29 +211,37 @@ internal class OrchestratorViewModel @Inject constructor(
...
     ) {
         if (currentStep.id == StepId.FACE_CAPTURE && result is FaceCaptureResult) {
-            val matchingStep = steps.firstOrNull { it.id == StepId.FACE_MATCHER }
+            val captureParams = currentStep.params?.let { it as? FaceCaptureParams }
+            val matchingStep = steps.firstOrNull { step ->
+                if (step.id != StepId.FACE_MATCHER) {
+                    false
+                } else {
+                    val stepSdk = step.params?.let { it as? MatchStepStubPayload }?.faceSDK
+                    stepSdk == captureParams?.faceSDK
```
BurningAXE, line 220: Interesting - had we completely missed this part previously?

luhmirin-s: I was as confused when I noticed that. It probably never came out since we did not have 2 Face SDKs to be enabled at the same time before. Even now, this could ever be an issue in the dev build if you enable ROC and SimFace.

PR #1275, opened 2025-07-21
[MS-1076] Fix: SID is not handling the enrolment record events when there is the metadata field

alexandr-simprints:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1076)

Jackson attempted to resolve the subtype for the `metadata` field. Because this field is within a polymorphic class (`ApiBiometricReference`), deserialization failed with a missing type error.

First known affected version: **2025.2.0 (testing phase)**

* Describe how the issue was found and what causes it
```Could not resolve subtype of [map type; class java.util.HashMap, [simple type, class java.lang.String] -> [simple type, class java.lang.String]]: missing type id property 'type' (for POJO property 'metadata')```
It is caused by the `Jackson` library applying polymorphic deserialization to the `metadata` field. This property is inherited from a base class using a `@JsonTypeInfo` annotation.
Note: added comments to change the implementation to the one suggested here: https://github.com/FasterXML/jackson-databind/issues/3877#issuecomment-1508508854 
Current MinSDK prevents us from updating Jackson, hence the current implementation with explicit annotation in the subclasses

### Testing guidance

* Describe how the reviewers can verify that issue is fixed
Use dev environment, and login into the `m0Zk7c28tbpMkdTcIBm8` project. This project contains the `8eaee786-f749-4059-872a-9fc3ff454c42` event, containing the metadate, and it should be parsed correctly (keep an eye on logs, as the crash doesn't stop the application from running)

### Additional work checklist

### DIFF
--- a/infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/remote/models/face/ApiFaceReference.kt
+++ b/infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/remote/models/face/ApiFaceReference.kt
@@ -1,6 +1,7 @@
 package com.simprints.infra.enrolment.records.repository.remote.models.face
 
 import androidx.annotation.Keep
+import com.fasterxml.jackson.annotation.JsonTypeInfo
 import com.simprints.core.domain.face.FaceSample
 import com.simprints.core.tools.utils.EncodingUtils
 import com.simprints.infra.enrolment.records.repository.remote.models.ApiBiometricReference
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: infra/enrolment-records/repository/src/main/java/com/simprints/infra/enrolment/records/repository/remote/models/face/ApiFaceReference.kt [hunk 1]
```diff
@@ -1,15 +1,22 @@
...
 import com.simprints.infra.enrolment.records.repository.remote.models.ApiBiometricReference
 
 @Keep
+@ExcludedFromGeneratedTestCoverageReports("API model")
 internal data class ApiFaceReference(
     val id: String,
     val templates: List<ApiFaceTemplate>,
     val format: String,
+    // [MS-1076] The parent 'ApiBiometricReference' class should have its JsonSubTypes annotation updated to
+    // @JsonSubTypes.Type([...], looseHandling = true) once we update to SDK => 25 and Jackson => 2.16.0.
```
luhmirin-s, line 17: > once we update to SDK => 25

if :D

luhmirin-s (approved):

PR #1282, opened 2025-07-24
[MS-939] Sync UI/UX update

alex-vt:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-939)

### Notable changes

* The new sync UI/UX is visually unified across Settings screen, Dashboard card and Logout screen, while maintaining the relevant features revevant to these specific places.
* Code: shared common state & logic of sync UI/UX in Settings screen, Dashboard card and Logout screen.
* Code: UI unified state LiveData calculated from combined flows of data from data layer, instead of procedurally updating LiveData backing fields.
* Code: Sync UI view differences between Settings screen, Dashboard card and Logout screen configured by XML attributes.
* Image sync control.
* Automatic device config updates instead of menu option.
* Downsync disabled where not applicable.
* Sync progress bar appears "glowing"-animated even when not moving for a while, indicating ongoing progress.
* UI height changes / vertical jumps avoided when transitioning between sync progress stages.
* UI (card) width limited on wide devices for proper look and readability.

### Additional work checklist

* [x] Documentation update: ticket created
* [x] Test cases in Testiny are update: ticket created


feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/LogoutSyncViewModel.kt
```diff
@@ -2,20 +2,51 @@ package com.simprints.feature.dashboard.logout
...
+            projectId.isEmpty()
+        }.distinctUntilChanged().map { /* Unit on every "true" */ }.asLiveData()
+
+    val isLogoutWithoutSyncVisibleLiveData: LiveData<Boolean> = combine(
+        eventSyncManager.getLastSyncState(useDefaultValue = true).asFlow(),
+        syncOrchestrator.observeImageSyncStatus(),
+        configManager.watchProjectConfiguration(),
+        configManager.watchDeviceConfiguration(),
+    ) { eventSyncState, imageSyncStatus, projectConfig, deviceConfig ->
+        val isModuleSelectionRequired =
```
BurningAXE, line 45: How is module selection relevant to logout syncing?

feature/dashboard/src/main/res/layout/item_module_count.xml
```diff
@@ -3,25 +3,40 @@
...
-        android:padding="5dp"
-        tools:text="10" />
+        style="@style/Text.Body1.Secondary"
+        android:layout_width="0dp"
+        android:layout_height="32dp"
+        android:layout_gravity="center"
+        android:layout_marginEnd="7dp"
+        android:layout_weight="0.15"
+        android:gravity="end|center_vertical"
+        tools:text="39"
```
BurningAXE, line 38: Very specific 😄

feature/dashboard/src/main/res/values/attrs.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <declare-styleable name="FragmentContainerView">
```
BurningAXE, line 3: Maybe name this `SyncFragmentContainerView` to avoid clash with any other fragment container we might choose to parametrize in the future.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
         applySystemBarInsets(view)
 
         binding.selectedModulesView.adapter = moduleCountAdapter
+
         setupClickListeners()
         observeUI()
 
-        findNavController().handleResult<LoginResult>(
+        findNavController().handleResult(
```
BurningAXE, line 64: Shouldn't we keep the generic result type?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
-        binding.syncInfoToolbar.setNavigationOnClickListener {
-            findNavController().popBackStack()
+        binding.textEventSyncInstructionsNoModules.setOnClickListener {
+            findNavController().navigate(R.id.moduleSelectionFragment)
         }
-        binding.syncInfoToolbar.setOnMenuItemClickListener {
-            viewModel.refreshInformation()
-            true
+        binding.syncSettingsButton.setOnClickListener {
+            findNavController().navigate(R.id.syncInfoFragment)
```
BurningAXE, line 80: Aren't we on the same screen?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
-        }
-        viewModel.isSyncAvailable.observe(viewLifecycleOwner) {
-            binding.syncButton.isEnabled = it
+    private fun observeUI() {
+        renderSyncInfo(SyncInfo(), syncInfoConfig)
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                renderSyncInfo(SyncInfo(), syncInfoConfig)
+                viewModel.syncInfoLiveData.observe(viewLifecycleOwner) { syncInfo ->
+                    renderSyncInfo(syncInfo, syncInfoConfig)
```
BurningAXE, line 121: Wouldn't this setup lead to needless renderSyncInfo() calls?
- when Fragment is created initially we'll have:
renderSyncInfo(SyncInfo(), syncInfoConfig)
renderSyncInfo(SyncInfo(), syncInfoConfig)
renderSyncInfo(syncInfo, syncInfoConfig)
- when Fragment is paused and then resumed:
renderSyncInfo(SyncInfo(), syncInfoConfig)
renderSyncInfo(syncInfo, syncInfoConfig)

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -50,193 +57,358 @@ internal class SyncInfoViewModel @Inject constructor(
...
+        if (isReadyToLogOut) {
+            // "flick" the logout event to prevent persistence while not observed, and to avoid unexpected logout later
+            emit(Unit)
+            emit(null)
+        }
+    }.asLiveData()
+
+    val syncInfoLiveData: LiveData<SyncInfo> = combine8(
+        connectivityTracker.observeIsConnected().asFlow(),
+        authStore.watchSignedInProjectId().map(String::isNotEmpty),
```
BurningAXE, line 92: In relation to all the new watch* functions - observe is more idiomatic to Kotlin/Android and I would prefer to stick to that.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -50,193 +57,358 @@ internal class SyncInfoViewModel @Inject constructor(
...
+        eventSyncStateFlow,
+        imageSyncStatusFlow,
+    ) { eventSyncState, imageSyncStatus ->
+        val isReadyToLogOut =
+            isPreLogoutUpSync && eventSyncState.isSyncCompleted() && !imageSyncStatus.isSyncing
+                && !configManager.isModuleSelectionRequired()
+        return@combine isReadyToLogOut
+    }.debounce(LOGOUT_DELAY_MILLIS).transformLatest { isReadyToLogOut ->
+        if (isReadyToLogOut) {
+            // "flick" the logout event to prevent persistence while not observed, and to avoid unexpected logout later
```
BurningAXE, line 84: We have the LiveDataEvent class to solve this problem.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -50,193 +57,358 @@ internal class SyncInfoViewModel @Inject constructor(
...
-        if (eventSyncState != lastKnownEventSyncState) {
-            if (eventSyncState.isSyncCompleted() && eventSyncState.isSyncReporterCompleted()) {
-                load()
+    fun toggleImageSync() {
+        viewModelScope.launch {
+            val isImageSyncing = imageSyncStatusFlow.firstOrNull()?.isSyncing == true
+            if (isImageSyncing) {
+                syncOrchestrator.stopImageSync()
+            } else {
+                syncOrchestrator.startImageSync()
```
BurningAXE, line 323: Isn't `startImageSync` calling `stopImageSync` internally?

infra/event-sync/src/test/java/com/simprints/infra/eventsync/EventSyncManagerTest.kt
```diff
@@ -149,6 +170,39 @@ internal class EventSyncManagerTest {
         assertThat(result).isEqualTo(DownSyncCounts(26, isLowerBound = true))
     }
 
+    @Test
+    fun `countEventsToDownload bypasses cache when within max age`() = runTest {
```
BurningAXE, line 174: I think this and the next test have swapped titles!?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -50,193 +57,358 @@ internal class SyncInfoViewModel @Inject constructor(
...
+    private val imageSyncStatusFlow =
+        syncOrchestrator.observeImageSyncStatus()
+
+    val logoutEventLiveData: LiveData<Unit?> = combine(
+        eventSyncStateFlow,
+        imageSyncStatusFlow,
+    ) { eventSyncState, imageSyncStatus ->
+        val isReadyToLogOut =
+            isPreLogoutUpSync && eventSyncState.isSyncCompleted() && !imageSyncStatus.isSyncing
+                && !configManager.isModuleSelectionRequired()
```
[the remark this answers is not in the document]
BurningAXE: Unless I'm missing something, module selection is not relevant to logout syncing. When logging out we need to just upsync local data and selected modules are used just for downsyncing. You may have been fooled by the previous implementation which needlessly did downsync and hence needed modules to be selected.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/LogoutSyncViewModel.kt
alex-vt: As also discussed below, we are going to omit this requirement for the logout version of the UI, I'll update this.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
alex-vt: The `handleLoginResult` handler function has this argument type, so it's redundant here. Just a similar thing as for example here in the [ConnectScannerControllerFragment](https://github.com/Simprints/Android-Simprints-ID/blob/706a03a97cb998952a7ece585f133c42a8febdf8/fingerprint/connect/src/main/java/com/simprints/fingerprint/connect/screens/controller/ConnectScannerControllerFragment.kt#L115).

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
alex-vt: The LiveData<SyncInfo> here can take a perceptible amount of time before it emits the first value during the first `onCreate`..`onResume` sequence. While we are waiting for the first data, the views are in a loading / placeholder visual state. But still, depending on where in the app we are showing the sync info, the UI sections are somewhat different - that is configured in `syncInfoConfig`. By rendering immediately, we ensure the correct UI sections visible while the data still loads.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
alex-vt: I've tried using it first but if I remember correctly it was still "holding" on the event while the user was in a different screen (thus abandoning the logout), only to suddenly be consumed by the view immediately once the user returns, logging them out. It could be something else so I'll try `LiveDataEvent` again here - if it actually works well here I'll update.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
alex-vt: I see. The 
```
lifecycleScope.launch {
    repeatOnLifecycle(Lifecycle.State.RESUMED) {
```
shouldn't be slow. I think I still added the render before these to eliminate an initial UI section jump. I'll recheck and remove the line 116 if there's no difference indeed.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
+            minutesAgo < 2 * 24 * 60 -> getString(IDR.string.sync_info_footer_time_1_day)
+            else -> getString(IDR.string.sync_info_footer_time_days, minutesAgo / 60 / 24)
         }
 
-        val totalRecordsEntry = ModuleCount(
-            getString(IDR.string.dashboard_sync_info_total_records),
-            moduleCounts.sumOf { it.count },
-        )
-        moduleCountsArray.add(TOTAL_RECORDS_INDEX, totalRecordsEntry)
+    private fun View.setPulseAnimation(isEnabled: Boolean) {
```
luhmirin-s, line 372: Could be worth extracting into the `View.ext.kt` in ui-base or core.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/view/ConfigurableSyncInfoFragmentContainer.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.simprints.feature.dashboard.R
+import com.simprints.feature.dashboard.settings.syncinfo.SyncInfoFragmentConfig
+
+class ConfigurableSyncInfoFragmentContainer @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : FrameLayout(context, attrs, defStyleAttr) {
+
+    val syncInfoFragmentConfig: SyncInfoFragmentConfig? = attrs?.let {
```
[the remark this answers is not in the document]
alex-vt: It allows writing the view layouts for slightly different appearances of sync info in different places in the app. 

For example:
```xml
<com.simprints.feature.dashboard.view.ConfigurableSyncInfoFragmentContainer
        app:isSyncInfoStatusHeaderSettingsButtonVisible="true"
        ...> 
        <androidx.fragment.app.FragmentContainerView
                android:name="com.simprints.feature.dashboard.settings.syncinfo.SyncInfoFragment"
                ...
```

This way, the specifics of the sync info representation are defined statically and in the layout layer.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
alex-vt: The new UX intentinally allows the user to stop image sync. The reason could be, having control over a network-hogging process on a slow network. It's easy to disable this later if we decide to.

feature/dashboard/src/main/res/values/attrs.xml
```diff
@@ -0,0 +1,13 @@
...
+        <attr name="isSyncInfoToolbarVisible" format="boolean" />
```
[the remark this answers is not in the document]
alex-vt: The `mode=standalone/embedded` would shift the burden on configuration elsewhere. We would end up with these definitions somewhere else rather than right in the layouts. I intentionally avoid clustering the individual properties into "modes". Instead, the individual properties describe the configuration directly. Less unnecessary abstraction.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
+                }
+            }.joinToString(separator = "\n")
+
+        textView.text = progressText
     }
 
-    private fun updateModuleCounts(moduleCounts: List<ModuleCount>) {
-        val moduleCountsArray = ArrayList<ModuleCount>().apply {
-            addAll(moduleCounts)
+    private fun formatLastSyncTime(minutesAgo: Int): String =
```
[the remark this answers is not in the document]
alex-vt: Probably I haven't noticed it. I'd still prefer to keep a small manual customization of having "Just now" instead of "0 <units_of_minimal_resolution>".

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +57,350 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
+
+        // Sync button
+        binding.buttonSyncImagesNow.isEnabled = images.isSyncButtonEnabled
+        binding.buttonSyncImagesNow.text = getString(
+            when {
+                images.isProgressVisible -> IDR.string.sync_info_button_images_sync_stop
+                else -> IDR.string.sync_info_button_sync_images
+            }
+        )
+        binding.buttonSyncImagesNow.backgroundTintList = ColorStateList(
```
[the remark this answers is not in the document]
alex-vt: This button state set is one of its kind, so this gets it done. I'll probably move it to an XML selector.

infra/sync/src/main/java/com/simprints/infra/sync/SyncOrchestratorImpl.kt
```diff
@@ -137,6 +140,52 @@ internal class SyncOrchestratorImpl @Inject constructor(
...
+                val currentIndex = workInfos.firstOrNull()?.progress
+                    ?.getInt(SyncConstants.PROGRESS_CURRENT, 0)?.coerceAtLeast(0) ?: 0
+                val totalCount = workInfos.firstOrNull()?.progress
+                    ?.getInt(SyncConstants.PROGRESS_MAX, 0)?.takeIf { it >= 1 }
+                val progress = totalCount?.let { currentIndex to totalCount }
+                ImageSyncStatus(isSyncing, progress, secondsSinceLastUpdate)
+            }
+    }
+
+    private fun Flow<List<WorkInfo>>.associateWithIfSyncing() = transformLatest { workInfos ->
```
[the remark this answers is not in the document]
alex-vt: There's an edge case in the implementation, where a double emission is used, that's why it's a transformation.

A comment on this was requested above already. It will be something like:
```
    /**
     * Converts the flow of WorkInfo in the receiver into a flow of WorkInfo paired to whether sync is ongoing or not.
     *
     * Whether sync is ongoing or not - is calculated from the WorkInfo.
     * A special case is handled for a job that succeeds promptly: a "pulse" of positive sync is emitted additionally.
     * This allows immediately succeeding syncs to be detected in the return flow.
     */
```

infra/sync/src/main/java/com/simprints/infra/sync/SyncOrchestratorImpl.kt
```diff
@@ -137,6 +140,52 @@ internal class SyncOrchestratorImpl @Inject constructor(
...
+    override fun stopImageSync() {
+        workManager.cancelWorkers(SyncConstants.FILE_UP_SYNC_WORK_NAME)
+    }
+
+    override fun observeImageSyncStatus(): Flow<ImageSyncStatus> {
+        return workManager
+            .getWorkInfosFlow(WorkQuery.fromUniqueWorkNames(SyncConstants.FILE_UP_SYNC_WORK_NAME))
+            .associateWithIfSyncing()
+            .map { (workInfos, isSyncing) ->
+                val secondsSinceLastUpdate = imageSyncTimestampProvider.getSecondsSinceLastImageSync()
```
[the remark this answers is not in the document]
alex-vt: The edge case mentioned above is the reason behind this more elaborate handling. I've seen skipping to `SUCCEEDED`. If that happens, while it's obvious that the sync happened (because succeeded), the lack of `RUNNING` causes the sync state not to be emitted. The view model wouldn't know the sync happened, and that would disrupt its logic.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
alex-vt: Exactly. As the first "user", I was finding it more startling to be logged out right after pressing No, than to back out of the logout screen and doing it again. `LiveDataEvent` will work if we change to the automatic logic, let's do it that way then and we can revisit this if needed in the future.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/sync/LogoutSyncFragment.kt
```diff
@@ -63,38 +40,23 @@ class LogoutSyncFragment : Fragment(R.layout.fragment_logout_sync) {
...
-        syncViewModel.syncCardLiveData.observe(viewLifecycleOwner) { state ->
-            val isLogoutButtonVisible = isLogoutButtonVisible(state)
-            logoutSyncCard.render(state)
-            logoutButton.isVisible = isLogoutButtonVisible
-            logoutWithoutSyncButton.isVisible = isLogoutButtonVisible.not()
-            logoutSyncInfo.isInvisible = isLogoutButtonVisible
+        lifecycleScope.launch {
+            repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                viewModel.isLogoutWithoutSyncVisibleLiveData.observe(viewLifecycleOwner) { isLogoutWithoutSyncVisible ->
+                    logoutSyncInfo.visibility = if (isLogoutWithoutSyncVisible) View.VISIBLE else View.INVISIBLE
```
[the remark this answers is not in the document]
luhmirin-s: For the regular case of visible/gone `isVisible` function is prefereble, IMO. Since pretty much all flags are usually "somethingSomethingVisible". "invisible" is an edge-case that warrants the explicit usage.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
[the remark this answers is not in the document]
luhmirin-s: The problem is that we would be forced to support it in all of the languages, most of which we do not know how the numbering grammar works. It would be much safer to leave that for the system, rather than make life harder for the delivery.

infra/sync/src/main/java/com/simprints/infra/sync/SyncOrchestratorImpl.kt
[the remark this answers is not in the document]
BurningAXE: What's the use case for this "pulse"?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
```diff
@@ -43,151 +62,311 @@ internal class SyncInfoFragment : Fragment(R.layout.fragment_sync_info) {
...
-                    this,
-                    R.id.action_syncInfoFragment_to_login,
-                    loginArgs,
+    }
+
+    private fun renderSyncInfo(
+        syncInfo: SyncInfo,
+        config: SyncInfoFragmentConfig,
+    ) {
+        // note: ".isGone = not" is preferred to ".isVisible =" below for non-ambiguity of the no-show state
```
luhmirin-s, line 151: > ".isGone = not" is preferred

This is not true if you have to reverse all of the conditions. `isVisible` behaviour is how the Android framework has handled view visibility from the very beginning; it is not ambiguous.

luhmirin-s (commented):
Noticed a couple more potential improvements.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoFragment.kt
alex-vt: At this point and context, the current solution is adequate.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCase.kt
```diff
@@ -0,0 +1,320 @@
...
+    private var cachedEventCountToDownload: DownSyncCounts? = null
+    private var cachedEventCountToDownloadTimestamp: Long = 0
+
+    private suspend fun countEventsToDownloadWithCaching(): DownSyncCounts {
+        val timeNowMs = timeHelper.now().ms
+        cachedEventCountToDownload
+            ?.takeIf {
+                timeNowMs - cachedEventCountToDownloadTimestamp < COUNT_EVENTS_CACHE_LIFESPAN_MILLIS
+            }?.let {
+                return it
```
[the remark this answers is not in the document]
alex-vt: Smart cast won't work here.

```kotlin
val currentlyCachedEventCountToDownload = cachedEventCountToDownload
if (
    currentlyCachedEventCountToDownload != null &&
    timeNowMs - cachedEventCountToDownloadTimestamp < COUNT_EVENTS_CACHE_LIFESPAN_MILLIS
) {
    return currentlyCachedEventCountToDownload
}
```

is quite verbose. I'd suggest to keep using the language as it currently allows, and we can migrate if the future changes actually arrive. Anyway `return@let it` trips the unit test, so we are protected in this particular potential case.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
luhmirin-s: > But we don't want this to happen, not even a single consumption. Instead we want to forget the event that happened while the user was away. 

This sounds more like a shared flow without replay than a LiveData to me. Are we using the right tool for the job here?

infra/core/src/main/java/com/simprints/core/tools/extentions/Flow.ext.kt
```diff
@@ -0,0 +1,53 @@
+package com.simprints.core.tools.extentions
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.scan
+
+fun <T1, T2, T3, T4, T5, T6, T7, T8, T9, R> combine9(
```
luhmirin-s, line 9: This is getting out of hand. Can you add a TODO comment/ticket to review ways to simplify this in future? Potentially, by doing some nested combining of parts of the state instead.

luhmirin-s (approved):
There are a couple of things that we should address for the sake of maintainability in the future, but none of those are blocking in my opinion.

I would suggest waiting for at least 1-2 additional approvals before merging, given the scale of the changes.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCase.kt
```diff
@@ -227,13 +236,15 @@ internal class ObserveSyncInfoUseCase @Inject constructor(
...
-                isConnected &&
+                isEventSyncConnectionSufficient &&
                 !eventSyncState.isSyncFailed() &&
                 !eventSyncState.isSyncInProgress() &&
                 !isPreLogoutUpSync,
-            isInstructionNoModulesVisible = isConnected && isModuleSelectionRequired && !isEventSyncInProgress,
-            isInstructionOfflineVisible = !isConnected,
-            isInstructionErrorVisible = isConnected && eventSyncState.isSyncFailed(),
+            isInstructionCommCarePermissionVisible = isCommCarePermissionMissing,
+            isInstructionNoModulesVisible = isEventSyncConnectionSufficient && !isCommCarePermissionMissing && isModuleSelectionRequired
```
BurningAXE: Why is `isCommCarePermissionMissing` part of the logic for `isInstructionNoModulesVisible`?

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCase.kt
```diff
@@ -329,3 +329,17 @@ internal class ObserveSyncInfoUseCase @Inject constructor(
         private const val COUNT_EVENTS_CACHE_LIFESPAN_MILLIS = 10 * 1000L
     }
 }
+
+private sealed class EventSyncVisibleSection
```
BurningAXE, line 333: My comment about naming is actually more relevant here as the var name comes from the class name.

feature/dashboard/src/test/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCaseTest.kt
```diff
@@ -950,6 +960,75 @@ class ObserveSyncInfoUseCaseTest {
...
+        createUseCase()
+
+        val result = useCase().first()
+
+        assertThat(result.syncInfoSectionRecords.isInstructionCommCarePermissionVisible).isFalse()
+        assertThat(result.syncInfoSectionRecords.isInstructionDefaultVisible).isTrue()
+    }
+
+    @Test
+    fun `should disable sync button when CommCare permission is missing`() = runTest {
```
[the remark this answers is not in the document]
alex-vt: Makes sense! Updated the sync availability logic.

infra/core/src/main/java/com/simprints/core/tools/extentions/Flow.ext.kt
luhmirin-s: As I understand, it doesn't go over 5 because it becomes harder. While there is nothing inherently wrong with combining 9 flows in one large function, there should be a way to split it into several smaller (more focused) sub-state combinations purely for maintainability.

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCase.kt
```diff
@@ -169,6 +170,12 @@ internal class ObserveSyncInfoUseCase @Inject constructor(
             else -> OnStandby
         }
 
+        val isUpSyncStillAllowedWhenCommCareSyncBlockedByDeniedPermission = // quite a specific case hence the name
+            isCommCareSyncBlockedByDeniedPermission && projectConfig.canCoSyncData()
```
BurningAXE, line 175: Actually you can have CommCare "down" sync and Simprints up sync. Down and upsync configurations are independent of each other. Down can be **either** Simprints or CommCare while up can be any (or all) of Simprints and CoSync. However, CoSync upsync is a bit of a misnomer since data is not really upsynced to an endpoint but rather returned in Intent result.

feature/dashboard/src/test/java/com/simprints/feature/dashboard/settings/syncinfo/usecase/ObserveSyncInfoUseCaseTest.kt
```diff
@@ -1027,15 +1083,70 @@ class ObserveSyncInfoUseCaseTest {
...
+        every { any<ProjectConfiguration>().canSyncDataToSimprints() } returns false
+        createUseCase()
+
+        val result = useCase().first()
+
+        assertThat(result.syncInfoSectionRecords.isSyncButtonEnabled).isFalse()
+    }
+
+    @Test
+    fun `sync button should be disabled when only CommCare down-sync allowed but there is CoSync permission error`() = runTest {
```
BurningAXE, line 1142: **CommCare** permission error

PR #1295, opened 2025-08-04
[MS-1093] Implement bulk session event generation and insertion logic

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1093)

### Notable changes

This PR introduces the ability for developers and E2E tests to generate bulk session events using predefined SQL templates stored in the assets folder.
Instead of relying on the existing event repository (which maps events to DbEvent objects and was significantly slow—~1 minute for 15 sessions), this implementation uses raw SQL insertion templates. As a result, the same 15 sessions are now generated in approximately 450 milliseconds

### Testing guidance

* Check the module's README.md file for instructions on how to use the event generator and details on the available parameters.

### Additional work checklist

### DIFF
--- a/testing/data-generator/src/debug/assets/dummy_events/AUTHORIZATION.sql
+++ b/testing/data-generator/src/debug/assets/dummy_events/AUTHORIZATION.sql
@@ -1,0 +1,13 @@
+INSERT INTO
+  DbEvent
+VALUES
+  (
+    '__event_id__',
+    'AUTHORIZATION',
+    '__project_id__',
+    '__session_id__',
+    '{"id":"__event_id__","payload":{"createdAt":{"ms":1754153597998,"isTrustworthy":false},"eventVersion":2,"result":"AUTHORIZED","userInfo":{"projectId":"__project_id__","userId":{"className":"TokenizableString.Raw","value":"__module_id__"}},"type":"AUTHORIZATION"},"type":"AUTHORIZATION","scopeId":"__scope_id__","projectId":"__project_id__"}',
+    1754153597998,
+    0,
+    NULL
+  );
--- a/testing/data-generator/src/main/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCase.kt
+++ b/testing/data-generator/src/main/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCase.kt
@@ -1,0 +1,168 @@
+package com.simprints.feature.datagenerator.events
+
+import com.simprints.core.DispatcherIO
+import com.simprints.infra.events.EventRepository
+import com.simprints.infra.events.event.domain.models.scope.EventScopeType
+import com.simprints.infra.logging.Simber
+import kotlinx.coroutines.CoroutineDispatcher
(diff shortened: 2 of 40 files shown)


testing/data-generator/src/debug/assets/dummy_events/AUTHORIZATION.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO DbEvent VALUES('__event_id__','AUTHORIZATION','__project_id__','__session_id__','{"id":"__event_id__","payload":{"createdAt":{"ms":1754153597998,"isTrustworthy":false},"eventVersion":2,"result":"AUTHORIZED","userInfo":{"projectId":"__project_id__","userId":{"className":"TokenizableString.Raw","value":"__module_id__"}},"type":"AUTHORIZATION"},"type":"AUTHORIZATION","scopeId":"__scope_id__","projectId":"__project_id__"}',1754153597998,0,NULL);
```
luhmirin-s, line 1: Can we format SQL files for improved readability?

testing/data-generator/src/main/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCase.kt
```diff
@@ -0,0 +1,168 @@
...
+    }.flowOn(dispatcher)
+
+    private suspend fun insertEnrollmentEvents(
+        projectId: String,
+        moduleId: String,
+        attendantId: String,
+        enrolCount: Int,
+    ) {
+        batchInsertEvents(count = enrolCount) { scopeId ->
+            when (Random.nextInt(2)) {
```
luhmirin-s, line 58: It would be nice to have some control over the kind of records being generated for the 1:N pool test cases, since only the face records are relevant and having an unpredictable number of records in the pool will make the testing more complicated.

luhmirin-s (approved):
Added a couple of suggestions, otherwise looks great.

testing/data-generator/src/main/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCase.kt
meladRaouf: For the 1\:N case, I believe it's better to use the bulk records generator instead of the events generator. The records generator is faster, allows us to specify more details—such as the number of templates per record and their SDK types—and uses real templates, whereas the events generator relies on dummy data.
I am going to update the README to include the use cases for each feature

testing/data-generator/src/debug/assets/dummy_events/AUTHORIZATION.sql
meladRaouf: Yes this is good Idea.
Done

testing/data-generator/src/main/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCase.kt
meladRaouf: Updated the readme

PR #1305, opened 2025-08-07
Add AI exclusions file according to best practices

luhmirin-s:
* According to the Gemini doc, this file works similarly to gitignore, but for the AI code generators ingesting the contents of the repo.
* This is something we should have done before any bots on the code, but better later than never.


BurningAXE (approved):
Nice!

PR #1348, opened 2025-09-04
[MS-1061] Fix relogin issue

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1061)

First known affected version: **It’s been a long time since we started using Realm DB**

* Check the Jira ticket regarding the issue where, after a user logs out from one project and logs into another, Realm reuses the same db file, causing the user to lose some of the previously synced records.

### Notable changes

* Now, when logging out, SID closes the active DB connection, forces the creation of a new Realm file when relogin, and resolves the problem. As shown in this screenshot, two DB files are created—one for the old project and another for the new one—so there’s no need to delete the old file.
two db files

### Testing guidance

* Follow the steps n the jira ticket to make sure that the issue is resolved. 

### Additional work checklist

### DIFF
--- a/feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
+++ b/feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
@@ -1,6 +1,7 @@
 package com.simprints.feature.dashboard.logout.usecase
 
 import com.simprints.infra.authlogic.AuthManager
+import com.simprints.infra.enrolment.records.repository.EnrolmentRecordRepository
 import com.simprints.infra.enrolment.records.repository.local.migration.RealmToRoomMigrationFlagsStore
 import com.simprints.infra.sync.SyncOrchestrator
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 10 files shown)


alex-vt (approved):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
```diff
@@ -1,23 +1,28 @@
...
 internal class LogoutUseCase @Inject constructor(
     private val syncOrchestrator: SyncOrchestrator,
     private val authManager: AuthManager,
     private val flagsStore: RealmToRoomMigrationFlagsStore,
+    private val enrolmentRecordRepository: EnrolmentRecordRepository,
+    @DispatcherIO private val ioDispatcher: CoroutineDispatcher,
 ) {
-    operator fun invoke() = runBlocking {
-        // Cancel all background sync
+    operator fun invoke() = runBlocking(ioDispatcher) {
```
BurningAXE, line 19: Do we need to specify the dispatcher here given that Realm will switch anyway?

BurningAXE (approved):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/logout/usecase/LogoutUseCase.kt
meladRaouf: Yes this a good point, I removed the dispatcher

PR #1356, opened 2025-09-09
[MS-1145] Visual sync button responsiveness optimization

alex-vt:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1145)

First known affected version: 2025.3.0 dev only

* The forceEventSync and toggleImageSync invoke sync purely on the data layer, so the UI was remaining unaware of the forced sync command until data-driven evidence of sync was starting to appear. This could take seconds on slow devices.

### Notable changes

* Each forced sync, invoked by forceEventSync and toggleImageSync, immediately reshapes the flow of events: At first we immediately emit a sync state that is forcefully marked as in progress, for events or images separately. And we start ignoring sync states that happen before the true progress in data layer appears. Once the true progress in data layer starts, we keep showing that true progress. Additionally, an initial progress is emitted on start, before any forced sync invocations - to prevent getting stuck on start.

### Testing guidance

* Record and image sync buttons react immediately when used to start the sync.

### Additional work checklist


feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
```diff
@@ -75,16 +82,63 @@ internal class SyncInfoViewModel @Inject constructor(
...
+         *
+         * The problem: data layer-driven progress visualization is simple programmatically, but can be slow in the UI.
+         *
+         * How it would work without the optimization:
+         * The forceEventSync and toggleImageSync invoke sync purely on the data layer,
+         * so the UI may remain unaware of the forced sync command until data-driven evidence of sync starts appearing.
+         * This may take seconds on slow devices.
+         *
+         * How it works with the optimization:
+         * Each forced sync, invoked by forceEventSync and toggleImageSync, immediately reshapes the flow of events:
```
luhmirin-s, line 102: IMO, only this paragraph is relevant; the rest can be pruned.

Also, this is not a KDoc comment, so it should not have the double star at the beginning.

luhmirin-s (approved):

feature/dashboard/src/main/java/com/simprints/feature/dashboard/settings/syncinfo/SyncInfoViewModel.kt
alex-vt: Removed the double star, but let's stay a bit more descriptive than usual about this non-trivial feature. There's been a history of more comments requested in similar situations rather than less.

PR #1369, opened 2025-09-22
[MS-1165] MF-ID:  QR scanner screen

alexandr-simprints:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1165)
Will be released in: **2025.4.0**

#### QR Analyzer the from login screen is extracted to a common module in `ui-base` feature.
image

This allows the same QR Code analyzer and detector to be used on the Login screen, and
* [x] Login QR scanner tested, and confirmed to be working

#### QR Analyzer now has a `CropConfig` class that allows to specify what part of the full image to analyze.
This is a requirement for the Multi-Factor Identification workflow, as the HEW might have mutiple QRs in frame, and we would like to minimize the scanned area to a cut-out in the middle of the screen. `CropConfig` allows to pass coordinates of the cut-out rectangle within the screen, alongside device's orientation and image resolution.

image

### Additional work checklist


infra/ui-base/src/main/java/com/simprints/infra/uibase/camera/qrscan/CameraFocusManager.kt
```diff
@@ -1,4 +1,4 @@
-package com.simprints.feature.login.tools.camera
+package com.simprints.infra.uibase.camera.qrscan
```
luhmirin-s, line 1: I am a bit conflicted about this move. On one hand, it is an easy way to reuse without micro-modules. On the other hand, ui-base is added to every feature module, and that means that all of them get camera and QR handling added for no real reason.

Would be interested to hear others' opinions on this.

luhmirin-s (commented):
LGTM, left a couple of comments for your consideration. 

The only reason not to approve right away is the comment about the placement of QR code in the ui-base.

PR #1384, opened 2025-10-02
Track firebase sample uploads

luhmirin-s:

### Notable changes

* This was missed during the initial implementation - Firebase uploads should also be tracked in the same way the signed URL uploads are for later comparisons.

### Additional work checklist

### DIFF
--- a/infra/images/src/main/java/com/simprints/infra/images/remote/firebase/FirebaseSampleUploader.kt
+++ b/infra/images/src/main/java/com/simprints/infra/images/remote/firebase/FirebaseSampleUploader.kt
@@ -1,4 +1,4 @@
-package com.simprints.infra.images.remote.firestore
+package com.simprints.infra.images.remote.firebase
 
 import com.google.firebase.storage.FirebaseStorage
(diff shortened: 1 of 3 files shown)


infra/images/src/main/java/com/simprints/infra/images/remote/firestore/FirestoreSampleUploader.kt
```diff
@@ -17,10 +24,13 @@ import java.io.FileInputStream
 import javax.inject.Inject
 
 internal class FirestoreSampleUploader @Inject constructor(
```
ybourgery, line 26: It may be too late, but should this not be name `FirebaseSampleUploader` instead of Firestore as it's not uploading the sample to Firestore

luhmirin-s: Luckily, it does not affect many files, so I have renamed it for the sake of correctness. Tbh, I often forget that Firestore is not the same as FirebaseStorage 🤦🏻

ybourgery (approved):

PR #1606, opened 2026-03-04
[MS-1377] Remove graph init state persistence that was never called

BurningAXE:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1377)

First known affected version: **an old one**

### Notable changes

* Activity used the two parameter `onSaveInstanceState()` which is only called when `persistableMode="persistAcrossReboots"` flag is set
* It is not set for this activity and as a result this method was not never called
* This resulted in the `isGraphInitialized` flag not persisting between activity recreations
* In a surprising plot twist it turned out that this is actually the desired behaviour which explains why rotation was working as expected and it broke once I switched from the two parameter to the single parameter `onSaveInstanceState()` :D
* To avoid confusion and someone re-finding this issue and "fixing it", I'm deleting the dead code and changing the comment to state desired behaviour
* @alexandr-simprints since you wrote this initially, I'll wait for your review specifically before merging, in case I missed something

### Testing guidance

* Trigger any workflow
* Rotate the device
* Continue workflow

### Additional work checklist

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
@@ -2,7 +2,6 @@ package com.simprints.feature.orchestrator
 
 import android.content.Intent
 import android.os.Bundle
-import android.os.PersistableBundle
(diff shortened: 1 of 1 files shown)


luhmirin-s (approved):
Don't forget to test it :D

PR #1608, opened 2026-03-06
[MS-1379] Don't crash on huge metadata

BurningAXE:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1379)

First known affected version: **at least since the last navigation refactoring**

* Calling app can send in the metadata whatever extra info they want
* When this data is ~ 200K (!) we accept and pass it to OrchestratorFragment as argument
* OrchestratorFragment has 3 lazily initialized VM (loginCheck, clientApi, orchestrator)
* When app is minimized, each VM saves the fragment args in addition to a few more copies from the fragment and Nav controller
* This leads to the saved state Bundle exceeding the max size of 1MB
* *KABOOM*

### Notable changes

* Instead of doing risky changes and jumping through hoops, we simply don't pass intent data as fragment args but read it directly through the activity
* This avoids all the data copies on saving state while having no down sides

### Testing guidance

* Send an Intent with at least 200K metadata (I've made a modification in Intent Launcher that does this as otherwise copy-paste buffer cannot handle so much data)
* Wait for SID to open and process the request (e.g. reach the Consent screen)
* Minimize it
* Bring it back - should still be on the same screen which means it didn't crash

### Additional work checklist

### DIFF
--- a/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
+++ b/feature/orchestrator/src/main/java/com/simprints/feature/orchestrator/OrchestratorActivity.kt
@@ -2,12 +2,9 @@ package com.simprints.feature.orchestrator
 
 import android.content.Intent
 import android.os.Bundle
-import androidx.core.os.bundleOf
 import androidx.navigation.findNavController
 import com.simprints.core.tools.activity.BaseActivity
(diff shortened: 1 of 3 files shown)


alexandr-simprints (approved):
Don't see any problems with it at the first glance. What Intent Launcher version supports sending large amount of data to SID?

luhmirin-s (approved):

PR #1634, opened 2026-04-09
MS-1356 Better logout handling

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1356)

### Notable changes

* Piggy-backing on the freshly added broadcast mechanism to make the logout process more robust:
  * Data cleanup has been moved to a "expedited" background worker to decouple it from whichever UI screen is open in the moment. This also decouples it from device and project state workers and ensures that logout is processed even if sync workers are force-cancelled.
  * After the cleanup internal broadcast is sent. 
  * Any activity in the application handles the internal logout intent and processes it accordingly - orchestration finishes with error, dashboard navigates to login request screen.

Main reasons why `LogoutWorker` is part of `SyncOrchestrator` mechanism is that large part of its work is directly calling sync orchestrator.
The other logical place for it would be in auth modules, but that would introduce dependency loops and having a whole dedicated module for 2 small classes (30 lines total) felt like over-kill.

### Testing guidance

* Settings -> about -> logout -> wait for sync and it should log out
* Settings -> about -> logout -> force logout -> should log out
* Authenticate -> mark device compromised -> refresh configuration -> should log out with message
* Authenticate -> mark device compromised -> wait 15 minutes (on debug) -> should log out with message
* Authenticate -> mark device compromised -> start biometric flow -> wait 15+ minutes -> should log out with message
* Authenticate -> mark device compromised -> close app -> wait 15 minutes -> open app -> should be logged out

### Additional work checklist


infra/sync/src/main/java/com/simprints/infra/sync/logout/LogoutWorker.kt
```diff
@@ -0,0 +1,43 @@
...
+    override suspend fun doWork(): Result = withContext(dispatcher) {
+        crashlyticsLog("Started")
+
+        val isProjectEnded =
+            inputData.getBoolean(SyncConstants.LOGOUT_INPUT_IS_PROJECT_ENDED, false)
+        try {
+            logoutUseCase(isProjectEnded)
+            success()
+        } catch (t: Throwable) {
+            fail(t)
```
alexandr-simprints, line 40: Don't we want to log the details into Crashlytics here instead of relying on the upstream logging?

alexandr-simprints (approved):

infra/sync/src/main/java/com/simprints/infra/sync/logout/LogoutWorker.kt
luhmirin-s: `fail()` is a super-class method that already does the logging with the correct tags and prefixes and only then returns the worker result.

PR #1643, opened 2026-04-27
[MS-1421] Creating OCR framework to work with the OCR readouts with text filter and spatial constraints

alexandr-simprints:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1421)

### Notable changes

Adding a way of extracting text fields from scanned ID cards by combining text matching with spatial navigation. This is required because the non-credential fields (_name, date of birth, expiry date, etc._) cannot be reliably extracted using only pattern matching alone. The reader allows extraction rules to be expressed as human-readable queries:
```
// Spatial navigation
val dateOfBirth = reader.find {
    isBelow { containsText("date of birth") }
    matchesPattern(Regex("your-regex-pattern"))
    isAbove { containsText("expiry date") }
}

// Simple text matching pattern
val credential = ocrReader.find { matchesPattern(NHIS_PATTERN) }
```
#### Custom OCR models
ML Kit's `Text` class is converted into custom `OcrTextModel`. This keeps all extraction logic free of ML Kit dependencies and makes it possible to unit test.

#### Column-aware spatial filtering
Since the fields of interest on the documents that we're scanning are located in their respective columns, only `isBelow` and `isAbove` are used. However, there might be situations where two columns are located right next to each other, and in order to enforce horizontal column boundaries, a candidate line is only returned if its left edge falls within the horizontal bounds of the anchor. 

If a column-aware spatial filtering is not applied then for a document layout as below 
```
Date of birth            Date of expiry
11/11/1999                12/12/2028 
```
the following query
```
val dateOfBirth = reader.find {
    isBelow { containsText("date of birth") }
}
```
might equally return `11/11/1999` or `12/12/2028`. We want to avoid such scenarios, and explicitly check that the value's boundaries below field of interest (_date of birth_) start within the 'date of birth' width.

image

The drawing above demonstrates the logic for determining whether a line is within anchor's column. If the line's top-left `X` coordinate is between anchor's `X_left` and `X_right`, then the element is considered in the same column


feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
```diff
@@ -0,0 +1,43 @@
+package com.simprints.feature.externalcredential.screens.scanocr.reader
+
+import com.simprints.feature.externalcredential.model.BoundingBox
+
+/**
+ * Wrapper for all scanned text after the OCR
+ *
+ * @param blocks blocks of text, each containing multiple [OcrLine]
+ * @param allLines all lines from blocks sorted by top-right bounding box coordinates ascending
```
luhmirin-s: Does it sort horizontal-first or vertical-first?

Also, are those the same lines that appear in blocks or only lines that does not belong to any block?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
```diff
@@ -0,0 +1,27 @@
...
+ */
+class OcrReader(
+    val model: OcrText,
+) {
+    /**
+     * Executes the query defined in [block] and returns the first matching [OcrLine], or null.
+     * The [block] receives an [OcrQueryScope] as its receiver — call filter methods directly
+     * without any chaining or terminal call.
+     */
+    fun find(block: OcrQueryScope.() -> Unit): OcrLine? = OcrQueryScope(model.allLines).apply(block).find()
```
luhmirin-s: I think it might be better to have these above/below/toLeft/toRight constraints be more like list filters that help to narrow down on the specific item instead of returning the first matching item.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Entry point for querying an [OcrText].
+ *
+ * Usage:
+ * ```
+ * val model = OcrModelBuilder.build(mlKitText)
+ * val reader = OcrReader(model)
+ *
+ * val membershipNumber = reader.query {
+ *     matchesPattern(Regex("\\d{8}"))
```
luhmirin-s, line 11: Does the position of the terms in the scope affect the lookup in any way?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
alexandr-simprints: `allLines` is just a flat projection of every line from every block. 
- They are sorted top-to-bottom by their bounding box top coordinate
- vertical-first, no horizontal sorting

There are no lines outside of blocks. Every line belongs to a block, and `allLines` is simply a flattened array  of the same lines so the callers can query across the entire card without traversing the block hierarchy.

luhmirin-s: What if 2 lines are on the same vertical coordinate?

luhmirin-s: Bonus question - what if there are 2 separate lines on the roughly the same vertical line (like in columns), but the one further right and one pixel higher?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: The current design is specifically designed around the idea of how humans traverse cards - you look for the membership number, not a list of potential membership numbers. For instance, in case dates on Ethiopian ID card, we would just say '_take second line under the date of <something>_', which can then be expressed as : 
```
val firstDate = reader.find {
   isBelow { containsText("date of birth") }
}

// ‘Gregorian Calendar' date below Ethiopian (if was found)
val secondDate = firstDate?.let {
   reader.find { isBelow(firstDate) }
}
```
It seems that it's currently enough for our needs. Is there any other way you'd rather call for the query?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: No, the position of terms within the query block does not affect the result. 

All conditions are collected first and evaluated together against each line during the single pass:
```
return allLines.firstOrNull { line ->
    val isBelowAnchor = belowAnchor == null || (
        line.boundingBox.top > belowAnchor.boundingBox.top &&
        line.boundingBox.left < belowAnchor.boundingBox.right
    )
    val isAboveAnchor = aboveAnchor == null || (
        line.boundingBox.top < aboveAnchor.boundingBox.top &&
        line.boundingBox.left < aboveAnchor.boundingBox.right
    )
    filters.all { it(line) } && isBelowAnchor && isAboveAnchor
}
```
The order in which filters and spatial constraints are declared have no effect on which line is returned.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
alexandr-simprints: The lines are sorted by their vertical position (_`y` coordinate, ascending_)

> _[...] the one further right and one pixel higher?_

Logic checks if its `X_left` (_top left_) is between anchor's `X_left` and `X_right`. If yes, then it is picked. If not, then next line is checked for the same condition.

Image

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrQuery.kt
```diff
@@ -0,0 +1,49 @@
...
+    fun containsPattern(regex: Regex): OcrQuery = apply {
+        filters += { line -> regex.containsMatchIn(line.text) }
+    }
+
+    fun containsText(text: String): OcrQuery = apply {
+        filters += { line -> line.text.contains(text, ignoreCase = true) }
+    }
+
+    fun hasExactText(text: String): OcrQuery = apply {
+        filters += { line -> line.text.equals(text, ignoreCase = true) }
```
luhmirin-s, line 25: Nit-pick: if it is exact match, it should not ignore case :) Maybe it should be a flag with default value as well?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
```diff
@@ -0,0 +1,48 @@
...
+ */
+internal class OcrReader(
+    val ocrText: OcrText,
+) {
+    /**
+     * Executes the query defined in [block] and returns the first matching [OcrLine], or null.
+     * The [block] receives an [OcrQuery] as its receiver — call filter methods directly
+     * without any chaining or terminal call.
+     */
+    fun find(block: OcrQuery.() -> Unit): OcrLine? = runQuery(OcrQuery(::runQuery).apply(block))
```
luhmirin-s: `runQuery(OcrQuery(::runQuery).apply(block))` is hell of a line. 

Is there a way do the same without calling a method on an object that also takes the reference to the same method as an argument?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    private fun runQuery(scope: OcrQuery): OcrLine? {
+        val belowAnchor = scope.belowResolver?.invoke()
+        val aboveAnchor = scope.aboveResolver?.invoke()
```
luhmirin-s: I don't really see any recursion happening, am I safe assuming that this DSL only supports a single layer of nesting? If that is the case, it should be mentioned in the comments.

Or there should be something that would prevent me from writing `isBelow { isBelow{ isAbove { containsText("FooBar") }}}`, maybe there should be different interfaces for positional and textual queries/conditions and the positional scope can only contain a textual condition.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Wrapper for all scanned text after the OCR
+ *
+ * @param blocks blocks of text, each containing multiple [OcrLine]
+ * @param allLines all lines from blocks sorted by bounding box top coordinate ascending
+ */
+@ExcludedFromGeneratedTestCoverageReports("Data class")
+internal data class OcrText(
+    val blocks: List<OcrBlock>,
```
luhmirin-s: Are you using blocks anywhere? Seems to be a bit redundant.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrQuery.kt
alexandr-simprints: The thing with OCR is that it might confuse some letters for uppercase or lowercase. Think of `o` `O`, `k` `K`,`u` `U`, etc.

I think we would rather allow for case ignore in this scenario for a benefit of picking up the text from the image in the first place

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
alexandr-simprints: The coordinates of the parent block are used to display the zoomed image on a credential search screen
image

this way it's easier to confirm that the correct `label`+`value` pair of text is picked.

However, I've checked the code, and the `OcrBlock` object itself seems redundant at the moment. I will refactor to remove it

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: While possible, it is not really practical. The framework is built specifically for the field extraction of ID cards, where anchor resolution is always a simple text lookup. I don't see any real extraction rule where you would require more than one level of nesting. 

It seems that adding a constraint through separate interfaces would add code complexity, and it would outweight the benefit for the current use case.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: Yeah, I compressed it after multiple iterations, and it made sense back while I was at it :D 
I agree that there is a room for improvement, would something like this look better?

```
fun find(block: OcrQuery.() -> Unit): OcrLine? {
    val query = OcrQuery(::runQuery).apply(block)
    return runQuery(query)
}
```

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrQuery.kt
luhmirin-s: It is not "exact"then, no? :D Again, I am just semi-jokingly nitpicking here.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
luhmirin-s: Tbh, not really, it is spreading the confusion over more lines. 

Why do you need to pass the function reference **and** call the same function with the same object as argument?

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
luhmirin-s: The provided example is obviously not practical, but without explicit mention of this limitation I can imagine that somewhere in far future someone could try to add a nested rule like
```
isBelow { 
  isBelow { containsText("Some major title") } 
  containsText("some subtitle")
}
```
and be confused why it does not work as expected.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
luhmirin-s: I was only referring the the list of blocks in the `OcrText` object, since all operations seem to be done only on lines.

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: I've decoupled the `OcrQuery` execution from construction, and moved resolution logic into `OcrReader`. This makes the `OcrQuery` a plain data object that just holds the conditions.

its `find` method now looks more readable:
```fun find(block: OcrQuery.() -> Unit): OcrLine? = runQuery(OcrQuery().apply(block))```

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrModel.kt
alexandr-simprints: `OcrBlock` object appears to be redundant, removed it

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/scanocr/reader/OcrReader.kt
alexandr-simprints: KDocs updated. Preventing nesting at compile time adds a lot of unnecessary complexity, and I'd rather have a convention at this point. If that proves to be an issue later on, we can always refactor

luhmirin-s (approved):

PR #1647, opened 2026-04-30
Mock Random generator to always choose the first path in session event insertion test

meladRaouf:


luhmirin-s (approved):

testing/data-generator/src/test/java/com/simprints/feature/datagenerator/events/InsertSessionEventsUseCaseTest.kt
```diff
@@ -71,8 +74,8 @@ class InsertSessionEventsUseCaseTest {
         // Mock event insertion (suspend function)
         coEvery { mockEventRepository.executeRawEventInsertions(any()) } returns Unit
 
-        // Mock the Random generator to always choose the first path (ISO) in insertEnrollmentEvents
-        every { Random.nextInt(3) } returns 0
+        // Mock the Random generator to always choose the first path (ISO) in insertEnrollmentEvents and insertIdentifyEvents
+        every { mockRandom.nextInt(any()) } returns 0
```
BurningAXE, line 78: Maybe a stupid question as I'm not familiar with the test but shouldn't we test the full range of options?

meladRaouf: It would be excessive to test every possible flow for a tool whose purpose is simply to generate test data. The number of combinations grows quickly—three enrolment template types (ROC, NEC, and ISO for SimMatcher) multiplied by two scenarios (with or without external credentials)—resulting in many paths with limited added value.

BurningAXE (approved):

PR #1666, opened 2026-05-18
MS-1411 Use timestamp model for last location time

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1411)

### Notable changes

* As per last-minute suggestion, changing last location time from long to timestamp model.
* Also noticed that the time was not being sent to the API all together, fixed that as well.

### Testing guidance

* Run several sessions before installing the changes, do not up-sync data
* Install changes
* Run couple more sessions and trigger the upsync - there should be no errors

### Additional work checklist

### DIFF
--- a/feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
+++ b/feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import com.google.android.gms.location.LocationServices
 import com.google.android.gms.location.Priority
 import com.google.android.gms.tasks.CancellationTokenSource
+import com.simprints.core.tools.time.Timestamp
 import com.simprints.infra.config.store.ConfigRepository
 import com.simprints.infra.config.store.models.experimental
 import com.simprints.infra.events.event.domain.models.scope.Location
@@ -15,6 +16,8 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.tasks.await
 import javax.inject.Inject
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 
 internal class LocationManager @Inject constructor(
     @param:ApplicationContext val ctx: Context,
@@ -44,7 +47,11 @@ internal class LocationManager @Inject constructor(
                     Location(
                         latitude = lastLocation.latitude,
                         longitude = lastLocation.longitude,
-                        lastLocationTime = lastLocation.time,
+                        lastLocationTime = Timestamp(
+                            ms = lastLocation.time,
+                            isTrustworthy = false,
+                            msSinceBoot = lastLocation.elapsedRealtimeNanos.toDuration(DurationUnit.NANOSECONDS).inWholeMilliseconds,
+                        ),
                     ),
                 )
             }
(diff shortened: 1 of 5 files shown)


infra/events/src/main/java/com/simprints/infra/events/event/domain/models/scope/Location.kt
```diff
@@ -1,14 +1,15 @@
...
+import com.simprints.core.tools.time.Timestamp
 import kotlinx.serialization.Serializable
 
 @Keep
 @Serializable
 data class Location(
     var latitude: Double = 0.0,
     var longitude: Double = 0.0,
-    val lastLocationTime: Long? = null,
+    val lastLocationTime: Timestamp? = null,
```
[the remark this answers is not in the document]
luhmirin-s: This might break for developers, but the field was only added in this version so it should not affect any real users.

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
```diff
@@ -44,7 +47,11 @@ internal class LocationManager @Inject constructor(
                     Location(
                         latitude = lastLocation.latitude,
                         longitude = lastLocation.longitude,
-                        lastLocationTime = lastLocation.time,
+                        lastLocationTime = Timestamp(
+                            ms = lastLocation.time,
+                            isTrustworthy = false,
+                            msSinceBoot = lastLocation.elapsedRealtimeNanos.toDuration(DurationUnit.NANOSECONDS).inWholeMilliseconds,
```
BurningAXE, line 53: I guess the answer is yes but is that really the most efficient way to get it to milliseconds?

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
```diff
@@ -44,7 +47,11 @@ internal class LocationManager @Inject constructor(
                     Location(
                         latitude = lastLocation.latitude,
                         longitude = lastLocation.longitude,
-                        lastLocationTime = lastLocation.time,
+                        lastLocationTime = Timestamp(
+                            ms = lastLocation.time,
+                            isTrustworthy = false,
```
BurningAXE, line 52: Apart from being uniform with other timestamps, is there any value in using Timestamp if it's always going to be untrustworthy?

BurningAXE (approved):

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
luhmirin-s: As described by Etienne in the design doc comment - even if not fully trustworthy, the additional "msSinceBoot" can be useful and makes it more trustworthy than just a timestamp.

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationManager.kt
luhmirin-s: The most efficient would be to call `lastLocation.elapsedRealtime`, but that is only available from API30-something. 
Division by "NANOSECONDS_IN_MS" constant would also work, but this feels safer since I don't need to do the math myself. Performance difference is non-existent.

PR #1668, opened 2026-05-19
[MS-1036] Close ImageProxy even when preview is empty to avoid leaks

BurningAXE:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1036)
Will be released in: **2026.2.0** if considered safe, if not - will rebase to main and leave for next release

First known affected version: **2025.1.0**

* There's a small trickle of crashes from all versions since 2025.1.0 when preview is empty (during initialization or onStop/onStart). In that case we exit early but don't close the ImageProxy so it gets leaked. When all 4 slots get occupied - *KABOOM*

### Notable changes

* Fix is straightforward - just wrap the whole body in a use {} so that ImageProxy gets closed in any case. Should be perfectly safe, too.

### Testing guidance

* Probably hard to reproduce but open face preview, minimize and bring back to foreground a few times, rotate the device, etc.

### Additional work checklist

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
@@ -12,40 +12,39 @@ internal class CropToTargetOverlayAnalyzer(
     private val onImageCropped: (Bitmap) -> Unit,
 ) : ImageAnalysis.Analyzer {
     override fun analyze(image: ImageProxy) {
-        val previewRect = targetOverlay.circleRect
-        if (previewRect.isEmpty) return
+        val croppedBitmap = image.use {
+            val previewRect = targetOverlay.circleRect
+            if (previewRect.isEmpty) return
 
-        // Adjust overlay size to be fit-center with the image size
-        val scale = getSmallerRatio(
-            image.width,
-            image.height,
-            targetOverlay.width,
-            targetOverlay.height,
-        )
-        val scaledWidth = (targetOverlay.width * scale).toInt()
-        val scaledHeight = (targetOverlay.height * scale).toInt()
+            // Adjust overlay size to be fit-center with the image size
+            val scale = getSmallerRatio(
+                it.width,
+                it.height,
+                targetOverlay.width,
+                targetOverlay.height,
+            )
+            val scaledWidth = (targetOverlay.width * scale).toInt()
+            val scaledHeight = (targetOverlay.height * scale).toInt()
 
-        // Find the offsets caused by fit-center scaling
-        val offsetX = (max(image.width, scaledWidth) - min(image.width, scaledWidth)) / 2
-        val offsetY = (max(image.height, scaledHeight) - min(image.height, scaledHeight)) / 2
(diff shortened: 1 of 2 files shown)


face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
```diff
@@ -12,40 +12,40 @@ internal class CropToTargetOverlayAnalyzer(
...
+            val cropTop = offsetY + (previewRect.top * scale).toInt()
+            val cropHeight = (previewRect.height() * scale).toInt()
 
-        onImageCropped(
-            image.use {
+            onImageCropped(
                 Bitmap.createBitmap(
                     it.toBitmap(),
                     cropLeft,
                     cropTop,
```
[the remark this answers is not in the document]
luhmirin-s: This makes sense, IMO.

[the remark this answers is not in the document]
BurningAXE: Yep, I'll take the callback out.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
```diff
@@ -12,40 +12,40 @@ internal class CropToTargetOverlayAnalyzer(
     private val onImageCropped: (Bitmap) -> Unit,
 ) : ImageAnalysis.Analyzer {
     override fun analyze(image: ImageProxy) {
-        val previewRect = targetOverlay.circleRect
-        if (previewRect.isEmpty) return
```
luhmirin-s, line 16: While your option is definitely safer, wouldn't it be essentially the same if we just do
```
if (previewRect.isEmpty) {
  image.close()
  return
}
```

BurningAXE: That would be the surgical fix. I've refactored it to still employ one single use{} but call `onImageCropped()` after that.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
```diff
@@ -12,40 +12,39 @@ internal class CropToTargetOverlayAnalyzer(
...
+        val croppedBitmap = image.use {
+            val previewRect = targetOverlay.circleRect
+            if (previewRect.isEmpty) return
```
luhmirin-s, line 17: Does this `return` actually respect the `.use{}` contract? Shouldn't it be `return@use null` and than handling of null case?

luhmirin-s (approved):

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/CropToTargetOverlayAnalyzer.kt
BurningAXE: It does. `use{}` wraps the block in a `try{} finally{ close() }`

PR #1679, opened 2026-05-26
[MS-1430] Migrate models from Java Serializable to StepParams and StepResult interfaces

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1430)

### Notable changes

* Migrates several Kotlin model types from implementing java.io.Serializable directly  to using the project’s StepParams / StepResult marker interfaces, aligning these models with the step contract types used across the app.

### Testing guidance

* Normal flow navigation especially when "don't keep activities" setting is active 

### Additional work checklist

### DIFF
--- a/infra/core/src/main/java/com/simprints/core/domain/tokenization/TokenizableString.kt
+++ b/infra/core/src/main/java/com/simprints/core/domain/tokenization/TokenizableString.kt
@@ -1,13 +1,14 @@
 package com.simprints.core.domain.tokenization
 
 import androidx.annotation.Keep
+import com.simprints.core.domain.step.StepParams
+import com.simprints.core.domain.step.StepResult
 import com.simprints.core.domain.tokenization.TokenizableString.Raw
 import com.simprints.core.domain.tokenization.TokenizableString.Tokenized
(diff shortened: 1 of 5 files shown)


infra/core/src/main/java/com/simprints/core/domain/tokenization/TokenizableString.kt
```diff
@@ -21,7 +21,7 @@ import java.io.Serializable as JavaSerializable
 @Serializable(with = TokenizableStringSerializer::class)
 @OptIn(ExperimentalSerializationApi::class)
 @JsonClassDiscriminator("className")
-sealed class TokenizableString : JavaSerializable {
+sealed class TokenizableString : StepParams {
```
luhmirin-s: IIRC, it is also part of several result models so it should have the `StepResult` as well.

luhmirin-s (approved):

infra/core/src/main/java/com/simprints/core/domain/tokenization/TokenizableString.kt
BurningAXE: Is this leading to issues parsing the results? If yes, maybe we should patch it at least in 2026.2.0.

PR #1680, opened 2026-05-26
[MS-1444] Reset accumulators upon scanner disconnect to prevent lefto…

BurningAXE:
…ver bytes from corrupting parsing on next connection

[JIRA ticket](https://simprints.atlassian.net/browse/MS-1444)

First known affected version: **since forever**

* See ticket for reproduction steps
* What happens upon timeout is that the Accumulator's buffer is not cleared so there are some leftover bytes from the previous connection when a new one is established. This leads to parsing errors that then lead to consistent timeouts and reconnections. Since the whole thing is a singleton, the only way to recover is to kill the app.

### Notable changes

* Added an explicit reset in the Accumulator that is called upon disconnect. This should be safe since there is no use case where we want state preservation through reconnects.

### Testing guidance

* Test a few normal/happy paths
* Check out the ticket for reproduction steps of the bug

### Additional work checklist

### DIFF
--- a/fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v2/tools/accumulator/Accumulator.kt
+++ b/fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v2/tools/accumulator/Accumulator.kt
@@ -27,7 +27,7 @@ import kotlinx.coroutines.flow.flow
  * @param buildElementFromCompleteCollection Method with which to build the next available Element from the collection
  */
 abstract class Accumulator<in Fragment, in FragmentCollection, Element>(
-    initialFragmentCollection: FragmentCollection,
+    private val initialFragmentCollection: FragmentCollection,
     private val addFragmentToCollection: FragmentCollection.(Fragment) -> FragmentCollection,
     private val canComputeElementLengthFromCollection: (FragmentCollection) -> Boolean,
     private val computeElementLengthFromCollection: (FragmentCollection) -> Int,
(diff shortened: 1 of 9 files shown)


fingerprint/infra/scanner/src/main/java/com/simprints/fingerprint/infra/scanner/v2/tools/accumulator/Accumulator.kt
```diff
@@ -39,6 +39,16 @@ abstract class Accumulator<in Fragment, in FragmentCollection, Element>(
 
     private var currentElementLength: Int? = null
 
+    /**
+     * Resets the accumulator to its initial empty state. Should be called when the underlying
+     * stream is disconnected so that any partial element bytes do not leak into the next
+     * connection's stream and corrupt subsequent message parsing.
+     */
+    fun reset() {
+        fragmentCollection = initialFragmentCollection
```
luhmirin-s: Would it be possible to replace the `initialFragmentCollection` instance with a provider/builder function? 

I am not fully familiar with the inner workings of the arrays, but if there is a chance that `initialFragmentCollection` is used as a reference and mutated in place, then this does not actually reset anything. It would be safer to explicitly create a new instance on each reset.

luhmirin-s: Dug a bit deeper, and it seems that there is `Array.copy()` everywhere and the actual initial array is not modified, so it is fine. Although, provider function still would fit better with the rest of the API of the Accumulator.

BurningAXE: Yes, that would be safer, even though we most likely wouldn't touch that code ever again.

luhmirin-s (approved):

meladRaouf (approved):

PR #1696, opened 2026-06-04
Update VERSION_NAME to 2026.3.0

meladRaouf:


luhmirin-s (approved):
We don't really need this at the moment, but why not :D

PR #1697, opened 2026-06-04
Expand JaCoCo exclusion filters for generated code

meladRaouf:
This PR fixs the empty JaCoCo reports

### DIFF
--- a/build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
@@ -19,13 +19,43 @@ class PipelineJacocoConventionPlugin : Plugin<Project> {
         }
     }
 
-    private val fileFilter = listOf(
+    /**
+     * Unified exclusion list used for both class-directory filtering in the JacocoReport task
+     * and the JacocoTaskExtension excludes on each Test task.
+     */
+    private val excludedPatterns = listOf(
+        // Android generated
         "**/R.class",
         "**/R$*.class",
         "**/BuildConfig.*",
         "**/Manifest*.*",
+        // Test classes
         "**/*Test*.*",
+        // Android framework
         "android/**/*.*",
+        // JDK internals
+        "jdk.internal.*",
+        // Kotlin-generated classes — Jacoco cannot handle multiple "$" in class names
+        "**/*\$\$*",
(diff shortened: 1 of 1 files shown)


build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
```diff
@@ -35,16 +49,22 @@ class PipelineJacocoConventionPlugin : Plugin<Project> {
...
-            val javaTree = fileTree("${project.buildDir}/intermediates/javac/debug") {
-                include("**/classes/**")
-                exclude(fileFilter)
-            }
-            val kotlinTree = fileTree("${project.buildDir}") {
-                include("tmp/kotlin-classes/debug/**", "classes/kotlin/debug/**", "classes/kotlin/main/**")
-                exclude(fileFilter)
+            val compiledClassPaths = listOf(
+                "tmp/kotlin-classes/debug",
+                "intermediates/javac/debug/classes",
```
luhmirin-s, line 72: Does this support wildcards? We could just skip the whole "intermeriates/**"

build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
```diff
@@ -26,6 +26,20 @@ class PipelineJacocoConventionPlugin : Plugin<Project> {
         "**/Manifest*.*",
         "**/*Test*.*",
         "android/**/*.*",
+        "**/*Dagger*.*",
```
luhmirin-s, line 43: Do we need to do the same for any ksm/generaed code? For example retrofit.

build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
meladRaouf: I tried, but the coverage data disappeared from the XML report

build-logic/convention/src/main/kotlin/PipelineJacocoConventionPlugin.kt
meladRaouf: I added the Room DAOs, but I don't think we need to skip everything.

luhmirin-s (approved):

PR #1703, opened 2026-06-11
[MS-1459] Enhance version code computation and validation

meladRaouf:
<p class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text">This PR changes APK <span class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text" data-w-weight="semibold" data-w-inline="" data-w-whitespace="linebreaks">versionCode</span> generation to be derived from the app's <span class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text" data-w-weight="semibold" data-w-inline="" data-w-whitespace="linebreaks">versionName</span> instead of CI build order. The new scheme encodes the release year, minor version, patch version, and build sequence into a stable numeric versionCode.</p><p class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text">This fixes a downgrade issue that occurred when creating hotfix builds for older releases, where newly generated APKs could end up with a lower versionCode than versions already published on Google Play.</p><p class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text" data-w-weight="semibold">Example</p><div class="oVf2pG_TableContainer" data-w-column-sizing="equal" data-w-component="table-scroll">
versionName | versionCode
-- | --
2026.11.1 (run 157) | 261101157

</div><p class="zlAe0W_TextBase zlAe0W_Text" data-w-component="text">This ensures version codes remain predictable and monotonically increasing across rebuilds and hotfixes.</p>

### DIFF
--- a/.github/workflows/reusable-build-apk.yml
+++ b/.github/workflows/reusable-build-apk.yml
@@ -30,11 +30,8 @@ jobs:
         environment: ${{ inputs.build-environment }}
 
         env:
-            # VERSION_CODE formula: (year-2000)*10_000_000 + minor*100_000 + patch*1_000 + (run_number%1000)
+            # VERSION_CODE formula: (year-2000)*10_000_000 + minor*100_000 + patch*1_000 + run_number
             # Re-runs are disabled: github.run_attempt must be 1.
-            # Guarantees within one VERSION_NAME:
-            # - VERSION_CODE is unique only until github.run_number%1000 wraps, so collisions can occur every 1000 runs
-            # Required release policy: keep VERSION_NAME increasing and bump it before 1000 workflow runs for the same VERSION_NAME.
             VERSION_CODE: "set in lower step"
             VERSION_NAME: ${{ inputs.version-name }}
             VERSION_SUFFIX: ${{ inputs.build-environment }}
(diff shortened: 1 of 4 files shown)


luhmirin-s (commented):
Is there any way to "unit test" the computation? It looks correct, but I would feel safer to approve it if there were tests :D

.github/workflows/reusable-build-apk.yml
```diff
@@ -53,26 +50,33 @@ jobs:
...
+                    fi
+                    if [ $VERSION_MINOR -gt 99 ] || [ $VERSION_PATCH -gt 99 ]; then
+                      echo "ERROR: Minor ($VERSION_MINOR) and patch ($VERSION_PATCH) must each be 0-99"; exit 1
+                    fi
+                    RUN_SEQ=$(( ${{ github.run_number }} % 1000 ))
+                    VERSION_CODE_CALCULATED=$(( (VERSION_YEAR - 2000) * 10000000 + VERSION_MINOR * 100000 + VERSION_PATCH * 1000 + RUN_SEQ ))
+                    if [ $VERSION_CODE_CALCULATED -gt 2100000000 ]; then
+                      echo "ERROR: VERSION_CODE $VERSION_CODE_CALCULATED exceeds Android max 2100000000"; exit 1
+                    fi
+                    echo "VERSION_CODE=$(( (VERSION_YEAR - 2000) * 10000000 + VERSION_MINOR * 100000 + VERSION_PATCH * 1000 + RUN_SEQ ))" >> $GITHUB_ENV
```
BurningAXE: Why is this calculated again instead of using VERSION_CODE_CALCULATED from above?

BurningAXE (approved):
Looks good! Smart to trim the pipeline runs to 3 digits! 
I don't foresee any issues with this approach and it should be generous enough for lots of minor and patch releases.

BurningAXE:
> Is there any way to "unit test" the computation? It looks correct, but I would feel safer to approve it if there were tests :D

If we can unit test it, maybe we can put it on a test branch and feed it different values to see what it spits out.

.github/workflows/reusable-build-apk.yml
meladRaouf: Nice catch it  is indeed redundant will update it

meladRaouf:
> > Is there any way to "unit test" the computation? It looks correct, but I would feel safer to approve it if there were tests :D
> 
> If we can unit test it, maybe we can put it on a test branch and feed it different values to see what it spits out.

@luhmirin-s @BurningAXE 
I have extracted the version code generation into a bash script file and the added a unit test for it

I asked copilot to run tests for the version code generation 

> Is there any way to "unit test" the computation? It looks correct, but I would feel safer to approve it if there were tests :D

PR #1716, opened 2026-06-22
fix: upload session events in FIFO order

meladRaouf:
Same PR for release with minimum changes.


infra/events/src/main/java/com/simprints/infra/events/event/local/EventRoomDao.kt
```diff
@@ -10,13 +10,13 @@ import kotlinx.coroutines.flow.Flow
 
 @Dao
 internal interface EventRoomDao {
-    @Query("select * from DbEvent order by createdAt_unixMs desc")
+    @Query("select * from DbEvent order by createdAt_unixMs asc")
```
BurningAXE: Don't we need to reverse the order just on the scopes? Even order within the scope is irrelevant IMO so better to limit the changes.

meladRaouf: Yes you are right, I will revert this part in the release PR and keep it for the main branch PR

BurningAXE (approved):

PR #1723, opened 2026-06-25
[MS-1484] MFID: QR codes are no longer required to be confirmed. QR code readouts cannot be edited.

alexandr-simprints:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1484)
Will be released in: **2026.2.1**

### Notable changes

* Values of the QR code readouts can no longer be edited
* QR code readouts no longer require checkbox confirmation from the user 
image

The OCR readouts still have the 'edit' icon and the 'confirmation checkbox' in place
image

### Testing guidance

* Enable QR codes in the MFID settings
* Scan the QR code
* Make sure that you can recapture and continue
* Make that confirmation checkbox is hidden
* Make that the edit icon is hidden

### Additional work checklist

### DIFF
--- a/feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/search/model/SearchCredentialState.kt
+++ b/feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/search/model/SearchCredentialState.kt
@@ -3,6 +3,7 @@ package com.simprints.feature.externalcredential.screens.search.model
 import androidx.annotation.Keep
 import com.simprints.core.ExcludedFromGeneratedTestCoverageReports
 import com.simprints.core.domain.common.FlowType
+import com.simprints.core.domain.externalcredential.ExternalCredentialType
 import com.simprints.core.domain.tokenization.TokenizableString
 import com.simprints.feature.externalcredential.model.CredentialMatch
 import kotlin.Boolean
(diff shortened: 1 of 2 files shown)


feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/search/model/SearchCredentialState.kt
```diff
@@ -16,19 +17,31 @@ internal data class SearchCredentialState(
...
-            displayedCredential = scannedCredentialResult.credential,
-            flowType = flowType,
-            searchState = SearchState.Searching,
-            isConfirmed = false,
-            isEditingCredential = false,
-        )
+        ): SearchCredentialState {
+            // [MS-1484] Editing and confirming the readout value should only be available for the OCR readouts (i.e.: Ghana NHIS card)
+            // QR codes are not readable by humans, hence there is no point of asking the user to confirm the readout, nor to edit the readout
+            val isOcrReadout = when (scannedCredentialResult.credentialType) {
```
luhmirin-s, line 30: Will there be any other non-OCR options? Maybe it would be simpler to just do `val isOcrReadout = scannedCredentialResult.credentialType != ExternalCredentialType.QRCode`?

luhmirin-s (approved):

meladRaouf (approved):

feature/external-credential/src/main/java/com/simprints/feature/externalcredential/screens/search/model/SearchCredentialState.kt
alexandr-simprints: No other non-OCR options are considered in the foreseeable future as far as I can tell. I wanted to keep the explicit `when` for this specific case: if we agree to add a new credential type, we can refine the logic further, and not skip this condition.

luhmirin-s: What would be the benefit of that? To me it feels like simply adding more work in the future.

alexandr-simprints: 1. The explicit `when` makes sure we won't miss this case.
2. If we stick to `val isOcrReadout = credentialType != ExternalCredentialType.QRCode`, then, if non-OCR type is added, it is easy to miss this case since it is not explicit

Thus, I'd advise to sticking to the `when` option

PR #1735, opened 2026-07-07
MS-1491 Spoof check implementation

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1491)

[Design document](https://docs.google.com/document/d/1yOKwQL2V4Tv84HzFiWOoYEz0xAjk-Q9K-yUk1LCvX4w/edit?tab=t.0)

### Notable changes

* Added spoof checks when using ROC v3 face SDK (other SDKs have no-op stubs). Spoof check return the "probability that provided image is being spoofed" and must not be over the configured threshold. If any of the images are over the threshold and the mode is configured to be "ENFORCED", the capture is reset up to configured amount of times. 

https://github.com/user-attachments/assets/d6361e46-193e-443c-a570-0f37509ce725

* In "RECORDED" mode the spoof check is performed and results are stored, but the flow is not reset in case of failed check.

#### Custom config 

Will be added to the documentation before merging this PR.

```
"spoofCheckMode": "ENFORCED",  // Or "DISABLED", or "RECORDED"
"spoofCheckThreshold": 0.33,  // Probability cutoff in (0; 1) range
"spoofMaxAttempts": 2,  // How many attempts are users forced to do until all images pass
"spoofMaxBitmapSize": 1500,  // Largest image size to scale down to before check to ensure IOD is within range
"spoofMinValidationUiDurationMs": 2000,  // At least how long the processing is shown
"spoofMinValidationErrorUiDurationMs": 2000,  // At least how long error is shown
```

### Testing guidance

* Configure a project with face modality, ROC v3 and 2-5 samples to capture.
* Add the custom configuration `"spoofCheckMode": "ENFORCED"` and try to run identification flows:
  * with printed/screen faces - it should force re-capture;
  * with real faces - it should proceed without re-capture.
* Add the custom configuration `"spoofCheckMode": "RECORDED"` and try to run identification flows:
  * it should proceed without re-capture on any face;
  * spoof score is recorded in the session events (either in troubleshooting screen or network data).
* Currently the logging is a bit verbose to help with debugging, it will be cleared before merging the PR.

### Additional work checklist

### DIFF
--- a/infra/events/src/main/java/com/simprints/infra/events/event/domain/models/FaceCaptureEvent.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/event/domain/models/FaceCaptureEvent.kt
@@ -64,7 +64,7 @@ data class FaceCaptureEvent(
     ) : EventPayload() {
         override fun toSafeString(): String =
             "result: $result, attempt nr: $attemptNb, auto-capture: $isAutoCapture, fallback: $isFallback, " +
-                "quality: ${face?.quality},  format: ${face?.format}"
+                "quality: ${face?.quality},  format: ${face?.format}, spoof score: ${face?.spoofScore}"
 
         @Keep
         @Serializable
@@ -73,6 +73,8 @@ data class FaceCaptureEvent(
             var roll: Float,
             val quality: Float,
             val format: String,
+            val spoofScore: Float? = null,
+            val spoofSkipReason: SpoofSkipReason? = null,
         )
 
         @Keep
@@ -85,6 +87,14 @@ data class FaceCaptureEvent(
             OFF_ROLL,
             TOO_CLOSE,
             TOO_FAR,
+        }
+
+        @Keep
+        @Serializable
+        enum class SpoofSkipReason {
+            IMAGE_TOO_SMALL,
+            IOD_TOO_SMALL,
+            IOD_TOO_LARGE,
         }
     }
 
--- a/infra/events/src/main/java/com/simprints/infra/events/event/domain/models/FaceCaptureEvent.kt
+++ b/infra/events/src/main/java/com/simprints/infra/events/event/domain/models/FaceCaptureEvent.kt
@@ -87,6 +87,7 @@ data class FaceCaptureEvent(
             OFF_ROLL,
             TOO_CLOSE,
             TOO_FAR,
+            SPOOFED,
         }
 
         @Keep
--- a/face/capture/src/main/java/com/simprints/face/capture/models/FaceDetection.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/models/FaceDetection.kt
@@ -3,6 +3,7 @@ package com.simprints.face.capture.models
 import android.graphics.Bitmap
 import com.simprints.core.tools.time.Timestamp
 import com.simprints.face.infra.basebiosdk.detection.Face
+import com.simprints.face.infra.basebiosdk.detection.SpoofCheckResult
 import com.simprints.infra.images.model.SecuredImageRef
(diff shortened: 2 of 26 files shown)


infra/events/src/main/java/com/simprints/infra/events/event/domain/models/FaceCaptureEvent.kt
ybourgery, line 82: @luhmirin-s Don't you need to add the `SPOOFED` result too?

luhmirin-s: That is a good question, it was obvious in the initial version of the document where spoof check was part of the feedback loop. Spoof detection is now fully separate process and these result values are taken directly from the feedback loop statuses. 

We could add the value and assign it after the check, but I think it might be unnecessary since the score is already present in the face data,

ybourgery: But if the face is spoofed, it means that we would not use the FaceCapture and SID will not extract the template for it no?

ybourgery: And in this case what would be the result in the FaceCapture, `VALID` even if it's not valid?

luhmirin-s: Since the spoof check is a computation heavy operation (it takes tens of ms to extract template and couple hundred to spoof check), it is done after all "valid" captures are gathered. So the capture can be "valid" (there is a face in the right position), but still have a high probability of being spoofed.

ybourgery: Let's chat only in the design doc so that it would be easier for other people to look at it comment on it

luhmirin-s: Added the value. Only marking the captures in the event just before saving it to the database, so that it does not interfere with existing UI logic.

alexandr-simprints:
image

I know it might be subjective, but in my opinion, the 'Unable to verify' state is unclear:
- The color scheme does not differ from the 'Scanning phase'
- There might not be enough time to read the text in the field setting
- Lack of required user input. Automation is good but the 'Unable to verify' state is displayed for a static amount of time, and we cannot be sure whether it's too short or too long


My suggestion would be to add a popup bottom sheet dialog when the 'Unable to verify' state is reached. Considering that the spoof positives should be an edge case, I think it's best to give the user enough time to familiarize with the message, and confirm their intent to recapture

luhmirin-s:
> * The color scheme does not differ from the 'Scanning phase'
> * There might not be enough time to read the text in the field setting
> My suggestion would be to add a popup bottom sheet dialog when the 'Unable to verify' state is reached. Considering that the spoof positives should be an edge case, I think it's best to give the user enough time to familiarize with the message, and confirm their intent to recapture

I discussed this with Jon and we decided that the current state is "good enough" for a first version as we are not sure how prevalent this state will be how disruptive it should be. It is very likely that "enforced" mode would not be used in this version to gather some real-life data first.

> * Lack of required user input. Automation is good but the 'Unable to verify' state is displayed for a static amount of time, and we cannot be sure whether it's too short or too long

There are custom config fields to adjust UI timing specifically for this reason.

alexandr-simprints (approved):

meladRaouf (approved):

PR #1762, opened 2026-07-29
[MS-1520] Don't send empty moduleIds in DeviceConfigurationUpdatedEvent

BurningAXE:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1520)

First known affected version: **2026.2.0**

* If selected moduleIds on the device are empty, it generates an event with an empty array (instead of null) for moduleIds but doesn't list them in tokenized fields (as there's nothing tokenized in there). However, the backend validation trips as it sees a path that should be tokenized but is not listed as such.

### Notable changes

* When moduleIds are empty - set the value to null instead of empty array so that JSON conversion omits it

### Testing guidance

* Login with a project configured with module partitioning but don't select modules
* Go to settings and change language
* Sync
* Backend should not complain

### Additional work checklist

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/usecases/TokenizeEventPayloadFieldsUseCase.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/event/usecases/TokenizeEventPayloadFieldsUseCase.kt
@@ -19,15 +19,20 @@ internal class TokenizeEventPayloadFieldsUseCase @Inject constructor(
                 project = project,
             )
         }
-        val tokenizedListFields = event.getTokenizableListFields().mapValues { (tokenKeyType, listFieldValue) ->
-            listFieldValue.map { fieldValue ->
-                tokenizationProcessor.tokenizeIfNecessary(
-                    tokenizableString = fieldValue,
-                    tokenKeyType = tokenKeyType,
-                    project = project,
-                )
+        val tokenizedListFields = event
+            .getTokenizableListFields()
+            // Empty lists have nothing to tokenize, and propagating them would replace an absent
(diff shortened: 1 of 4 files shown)


infra/events/src/main/java/com/simprints/infra/events/event/domain/models/DeviceConfigurationUpdatedEvent.kt
```diff
@@ -70,10 +70,12 @@ data class DeviceConfigurationUpdatedEvent(
         TokenKeyType.ModuleId to payload.configuration.downSyncModules.orEmpty(),
     )
 
+    // An empty list must be normalised back to null, otherwise the field is serialised as an empty
+    // array that the backend expects to be listed as tokenized, which it never can be.
     override fun setTokenizedListFields(map: Map<TokenKeyType, List<TokenizableString>>): Event = this.copy(
         payload = payload.copy(
             configuration = payload.configuration.copy(
-                downSyncModules = map[TokenKeyType.ModuleId] ?: payload.configuration.downSyncModules,
+                downSyncModules = (map[TokenKeyType.ModuleId] ?: payload.configuration.downSyncModules)?.ifEmpty { null },
```
luhmirin-s: Wouldn't it be safer to fix it in `TokenizeEventPayloadFieldsUseCase:23` by filtering/skipping any fields with empty lists?

That way we would not need to remember to apply the same fix for any list field in any future events.

BurningAXE: Good idea! I'll rework.

luhmirin-s (approved):

alexandr-simprints (approved):

PR #1772, opened 2026-08-06
[MS-1418] Replace WorkManager location collection with in-process coroutine

meladRaouf:
skip it and focus on writing the PR description instead.

[MS-1418](https://simprints.atlassian.net/browse/MS-1418)

### Notable changes

* Replaced the WorkManager-based location collection with a lightweight in-process coroutine implementation.
* Added logging around the lifecycle of location collection (`Started` / `Saved ... (took Xms)` / `Finished` / `Stopped`) to make it easier to trace timing and diagnose issues in Crashlytics.
* Tested manually and confirmed the timing is nearly identical between the WorkManager-based approach and the new implementation.

### Testing guidance

* Start any flow and confirm the user's location is still saved into the session (check session location fields / logs tagged `CollectLocationUseCase`).
* Verify cancelling setup early (before location resolves) doesn't crash and logs `Stopped collecting location`.

### Additional work checklist

[MS-1418]: https://simprints.atlassian.net/browse/MS-1418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


luhmirin-s:
> Tested manually and confirmed the timing is nearly identical between the WorkManager-based approach and the new implementation.

If the timing is nearly identical, is there any good reason to change how it works already?

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationStoreImpl.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class LocationStoreImpl @Inject constructor(
+    private val collectLocation: CollectLocationUseCase,
+) : LocationStore {
+    private var collectionJob: Job? = null
+
+    override fun collectLocationInBackground() {
```
luhmirin-s, line 21: I think the whole point was to do the collection coroutine in the context/scope of the caller to get the benefit of structured concurrency and avoid tracking this job manually.

meladRaouf: My  Idea  was that the collectLocationInBackground gets started in the setup fragment and gets canceled from the orchestrator. That is why I needed ahigher context than the caller so I used AppScope. 
 I implemented it in a nicer way now in LocationStoreImpl

meladRaouf:
> > Tested manually and confirmed the timing is nearly identical between the WorkManager-based approach and the new implementation.
> 
> If the timing is nearly identical, is there any good reason to change how it works already?

There is definitely some overhead involved in creating and cancelling the WorkManager, but it’s not noticeable in practice.

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationStoreImpl.kt
BurningAXE: I think the expected result is for location collection to run as long as the session is active. If we switch it to cancel as soon as Setup is done, that's a behaviour change and likely a negative one.

meladRaouf: @BurningAXE The cancellation is happening as before, at the end of the orchestration fragment rather than during the setup.

luhmirin-s: @BurningAXE technically, it would be until location is received or session interrupted. This implementation seems to be slightly simpler to understand, although I am surprised that the time benefits are negligible.

feature/setup/src/main/java/com/simprints/feature/setup/location/CollectLocationUseCase.kt
```diff
@@ -0,0 +1,48 @@
...
+     * track and cancel a [kotlinx.coroutines.Job] manually.
+     */
+    suspend operator fun invoke() {
+        val requestStartTimeMs = System.currentTimeMillis()
+        Simber.i("Started collecting location", tag = TAG)
+        try {
+            locationManager
+                .requestLocation()
+                .filterNotNull()
+                .collect { location -> runCatching { saveUserLocation(location, requestStartTimeMs) } }
```
luhmirin-s: Wouldn't the `runCatching()` wrap and hide the cancellation exception, or any other error? Which means that the outer try-catch is useless.

feature/setup/src/main/java/com/simprints/feature/setup/location/LocationStoreImpl.kt
BurningAXE: I was commenting on the suggestion to use the scope of the caller which is the Setup VM. Would have been easier if we had a dedicated session scope. (We do have SessionCoroutineScope but that one is strictly reserved for events and I don't think this single use case merits creating a new one)

luhmirin-s (approved):

feature/setup/src/main/java/com/simprints/feature/setup/location/CollectLocationUseCase.kt
meladRaouf: I replaced it with an explicit try/catch that rethrows  CancellationException.

PR #1782, opened 2026-08-12
[MS-1541] Stop event sync retries on network connection failure

meladRaouf:
[MS-1541](https://simprints.atlassian.net/browse/MS-1541)

First known affected version: **2026.3.0**

* When the device is offline, `EventUpSyncTask.uploadEventScopeType` retries uploading closed event scopes in a `while` loop with no delay/backoff.
* Each failed upload attempt calls `authStore.getFirebaseToken()` under the hood, which throws `NetworkConnectionException` when `FirebaseAuthManager` can't reach the network. This is caught and unconditionally logs a new `EventUpSyncRequestEvent`, unlike the success path which only logs when there's useful content.
* Because the failed scope is never deleted, the loop's exit condition (`getClosedEventScopesCount(...) > 0`) stays true, so it retries immediately with no backoff — spinning indefinitely while offline and creating a new event on every iteration, which shows up as accumulating extra events on the dashboard sync screen.

### Changes

* `EventUpSyncTask.uploadEventScopeType`: when the upload fails with a `NetworkConnectionException`, break out of the retry loop for that scope type instead of retrying immediately, deferring the retry to the next scheduled sync attempt. This limits the failure to a single logged event per sync run while offline instead of an unbounded tight loop.

### Testing guidance

* To verify manually: put the device in airplane mode with pending events to sync, trigger a sync, and confirm the sync screen event count stops growing after a single failed attempt instead of climbing continuously.

### Additional work checklist

- [ ] Effect on other features and security has been considered
- [ ] Design document marked as "In development" (if applicable)
- [ ] External (Gitbook) and internal (Confluence) Documentation is up to date (or ticket created)
- [ ] Test cases in Testiny are up to date (or ticket created)
- [ ] Other teams notified about the changes (if applicable)

[MS-1541]: https://simprints.atlassian.net/browse/MS-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
+++ b/infra/event-sync/src/main/java/com/simprints/infra/eventsync/sync/up/tasks/EventUpSyncTask.kt
@@ -42,6 +42,7 @@ import com.simprints.infra.network.exceptions.BackendMaintenanceException
 import com.simprints.infra.network.exceptions.NetworkConnectionException
 import com.simprints.infra.network.exceptions.SyncCloudIntegrationException
 import com.simprints.infra.serialization.SimJson
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.FlowCollector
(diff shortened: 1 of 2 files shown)


luhmirin-s:
@meladRaouf can you please stack this on top of #1783? There was a significant rework and it might need slightly different approach.

meladRaouf:
> @meladRaouf can you please stack this on top of #1783? There was a significant rework and it might need slightly different approach.

I will be waiting for your PR to be merged

luhmirin-s (approved):

PR #1791, opened 2026-08-17
[MS-1545] Increment attempt number on face recapture

meladRaouf:
[MS-1545](https://simprints.atlassian.net/browse/MS-1545)

First known affected version: **v2021.3.0**

* Found while investigating why `FaceCaptureEvent.attemptNb` was always `0` in analytics, regardless of how many times a user retook their face capture.

### Notable changes

* `FaceCaptureViewModel.recapture()` now increments `attemptNumber` on every recapture (covers the confirmation screen's decline/back-press/retake actions and the "no face detected" auto-retry).
* `LiveFeedbackViewModel.showValidationErrorAndReset()` now also increments the attempt count on a failed spoof-check retry, so each retry gets a distinct `attemptNb` instead of collapsing into one.
* `attemptNumber` is now backed by `SavedStateHandle` so it survives process death, and encapsulated with a private setter (only mutable via `recapture()`/internal reset) instead of a public mutable field.
* `initFaceBioSdk` resets the per-step state (`attemptNumber`, `initialised`) when a new capture step starts for a different face SDK, since `FaceCaptureViewModel` is Activity-scoped and can be reused across multiple face-capture steps (e.g. a project configured with more than one face SDK).
* Removed a stale/misleading default (`LiveFeedbackViewModel.attemptNumber = 1`) that was always overwritten before use.

### Testing guidance

* Manually: run a face capture flow, decline/retake at the confirmation screen a few times, and confirm the synced `FaceCaptureEvent.attemptNb` increments per retry instead of staying at `0`.

image
image

### Additional work checklist

[MS-1545]: https://simprints.atlassian.net/browse/MS-1545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
@@ -55,9 +55,10 @@ internal class FaceCaptureViewModel @Inject constructor(
     private val saveLicenseCheckEvent: SaveLicenseCheckEventUseCase,
     private val shouldShowInstructions: ShouldShowInstructionsScreenUseCase,
     @param:DeviceID private val deviceID: String,
+    private val captureAttemptTracker: CaptureAttemptTracker,
 ) : ViewModel() {
-    // Updated in live feedback screen
-    var attemptNumber: Int = 0
+    val attemptNumber: Int
+        get() = captureAttemptTracker.attemptNumber
     var samplesToCapture = 1
     var initialised = false
     lateinit var bioSDK: ModalitySdkType
@@ -92,6 +93,10 @@ internal class FaceCaptureViewModel @Inject constructor(
         activity: Activity,
         sdk: ModalitySdkType,
     ) = viewModelScope.launch {
+        if (::bioSDK.isInitialized && bioSDK != sdk) {
+            resetForNewCaptureStep()
+        }
+
         if (initialised) {
             Simber.i("Face bio SDK already initialised", tag = FACE_CAPTURE)
             return@launch
@@ -209,6 +214,16 @@ internal class FaceCaptureViewModel @Inject constructor(
         Simber.i("Starting face recapture flow", tag = FACE_CAPTURE)
         faceDetections = listOf()
         _recaptureEvent.send()
+    }
+
+    /**
+     * Resets the state that is scoped to a single capture step
+     **/
+    private fun resetForNewCaptureStep() {
+        Simber.i("Resetting face capture state for a new capture step", tag = FACE_CAPTURE)
+        captureAttemptTracker.reset()
+        initialised = false
+        faceDetections = emptyList()
     }
 
     private fun saveFaceDetections() {
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
@@ -118,7 +118,10 @@ internal class LiveFeedbackFragment : Fragment(R.layout.fragment_live_feedback) 
         // Wait till the views gets its final size then init frame processor and setup the camera
         binding.faceCaptureCamera.post {
             if (view != null) {
-                vm.initCapture(mainVm.bioSDK, mainVm.samplesToCapture, mainVm.attemptNumber)
+                vm.initCapture(
+                    bioSdk = mainVm.bioSDK,
+                    samplesToCapture = mainVm.samplesToCapture,
+                )
             }
         }
 
--- a/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackViewModel.kt
+++ b/face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackViewModel.kt
@@ -10,6 +10,7 @@ import com.simprints.core.tools.time.TimeHelper
 import com.simprints.face.capture.models.FaceDetection
 import com.simprints.face.capture.models.FaceTarget
 import com.simprints.face.capture.models.SymmetricTarget
+import com.simprints.face.capture.screens.CaptureAttemptTracker
 import com.simprints.face.capture.usecases.GetSpoofCheckConfigurationUseCase
 import com.simprints.face.capture.usecases.IsUsingAutoCaptureUseCase
 import com.simprints.face.capture.usecases.SimpleCaptureEventReporter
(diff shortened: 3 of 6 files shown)


luhmirin-s (approved):
Comments are a bit excessive, IMO.

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -207,10 +217,21 @@ internal class FaceCaptureViewModel @Inject constructor(
 
     fun recapture() {
         Simber.i("Starting face recapture flow", tag = FACE_CAPTURE)
+        attemptNumber++
         faceDetections = listOf()
         _recaptureEvent.send()
     }
 
+    /**
+     ** Resets the state that is scoped to a single capture step
```
luhmirin-s: KDoc comments should be formatted:
```
/**
*
*/
```

luhmirin-s (approved):

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -207,10 +217,20 @@ internal class FaceCaptureViewModel @Inject constructor(
...
     }
 
+    /**
+     * Resets the state that is scoped to a single capture step
+     **/
+    private fun resetForNewCaptureStep() {
+        Simber.i("Resetting face capture state for a new capture step", tag = FACE_CAPTURE)
+        attemptNumber = 0
+        initialised = false
+    }
```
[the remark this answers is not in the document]
BurningAXE: @meladRaouf This comment seems correct!?

BurningAXE (commented):
With this implementation we have two counters in two VMs and two increment sites. This is error-prone and may lead to bugs when we modify recapture/failed attempts/etc. I suggest we keep a single counter in FaceCaptureViewModel and have the LiveFeedbackViewModel use it through accessors and callbacks. In addition, if it can be incremented in a single place when capture flow starts (regardless of what triggered it) - that would cover all future scenarios.

meladRaouf:
> With this implementation we have two counters in two VMs and two increment sites. This is error-prone and may lead to bugs when we modify recapture/failed attempts/etc. I suggest we keep a single counter in FaceCaptureViewModel and have the LiveFeedbackViewModel use it through accessors and callbacks. In addition, if it can be incremented in a single place when capture flow starts (regardless of what triggered it) - that would cover all future scenarios.

I wanted to simplify the changes, but your suggestion is better than the current implementation. I’ve implemented it.

luhmirin-s:
> I wanted to simplify the changes, but your suggestion is better than the current implementation. I’ve implemented it.

I might be a simplification, but is passing a hard reference for a VM to another VM a good idea? I would suggest using a completely separate "tracker" instance that both VMs share via injection.

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
```diff
@@ -117,7 +117,11 @@ internal class LiveFeedbackFragment : Fragment(R.layout.fragment_live_feedback)
         // Wait till the views gets its final size then init frame processor and setup the camera
         binding.faceCaptureCamera.post {
             if (view != null) {
-                vm.initCapture(mainVm.bioSDK, mainVm.samplesToCapture, mainVm.attemptNumber)
+                vm.initCapture(
+                    bioSdk = mainVm.bioSDK,
+                    samplesToCapture = mainVm.samplesToCapture,
+                    captureAttemptTracker = mainVm,
```
alexandr-simprints: Please, let's not pass VM references to another VMs. The counter needs to be extracted and properly injected

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -55,9 +56,15 @@ internal class FaceCaptureViewModel @Inject constructor(
     private val saveLicenseCheckEvent: SaveLicenseCheckEventUseCase,
     private val shouldShowInstructions: ShouldShowInstructionsScreenUseCase,
     @param:DeviceID private val deviceID: String,
-) : ViewModel() {
-    // Updated in live feedback screen
-    var attemptNumber: Int = 0
+    private val savedStateHandle: SavedStateHandle,
+) : ViewModel(),
+    CaptureAttemptTracker {
```
alexandr-simprints: Please extract the tracker to a separate instance

alexandr-simprints (changes requested):

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackFragment.kt
meladRaouf: Agreed, I have extracted the counter into its own class.

meladRaouf:
> > I wanted to simplify the changes, but your suggestion is better than the current implementation. I’ve implemented it.
> 
> I might be a simplification, but is passing a hard reference for a VM to another VM a good idea? I would suggest using a completely separate "tracker" instance that both VMs share via injection.

Agreed, I have extracted the counter into its own class.

luhmirin-s (approved):

alexandr-simprints (approved):

face/capture/src/main/java/com/simprints/face/capture/screens/livefeedback/LiveFeedbackViewModel.kt
```diff
@@ -382,7 +381,8 @@ internal class LiveFeedbackViewModel @Inject constructor(
         emit(phase = LiveFeedbackState.Phase.VALIDATION_FAILED)
         val duration = measureTimedValue {
             // Still track the capture attempt events for analytics and troubleshooting
-            sendCaptureEvents(attemptNumber)
+            sendCaptureEvents(captureAttemptTracker.attemptNumber)
+            captureAttemptTracker.onNewCaptureAttemptStarted()
```
BurningAXE: Sorry to drag this on but I had one more point in my previous comment - instead of incrementing the counter when capture ends (which can be in many places) can we put in on the single path when capture starts (startCapture())? This way if we add a third way capture can end/fail, there won't be risk of forgetting to call increment.
Only caveat is that reset will have to set -1 as increment will be called on the first attempt, too.

meladRaouf: For the second point. 
I’d prefer not to use -1 here, as it introduces a special value just to represent the initial state.

face/capture/src/main/java/com/simprints/face/capture/screens/FaceCaptureViewModel.kt
```diff
@@ -55,9 +55,10 @@ internal class FaceCaptureViewModel @Inject constructor(
     private val saveLicenseCheckEvent: SaveLicenseCheckEventUseCase,
     private val shouldShowInstructions: ShouldShowInstructionsScreenUseCase,
     @param:DeviceID private val deviceID: String,
+    private val captureAttemptTracker: CaptureAttemptTracker,
 ) : ViewModel() {
-    // Updated in live feedback screen
-    var attemptNumber: Int = 0
+    val attemptNumber: Int
```
BurningAXE, line 60: I think this is not needed anymore!?

PR #1825, opened 2026-09-08
[MS-1593] Include CI build number in the  release APK filename

meladRaouf:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1593)
Will be released in: **2027.1.0**

### Notable changes

* Update the Android release workflow to append the build number to the generated APK filename. This will make it easier to differentiate APK URLs between different builds of the same release.

### Testing guidance

* when running the workflow the apk name should include the github run number 

### Additional work checklist

### DIFF
--- a/.github/workflows/reusable-upload-to-internal-and-release.yml
+++ b/.github/workflows/reusable-upload-to-internal-and-release.yml
@@ -15,6 +15,10 @@ on:
                 description: "Android version code for the build (used to download the universal APK from Play)"
                 type: string
                 required: true
+            build-number:
+                description: "CI build number to append to the released APK filename (e.g. 173)"
+                type: string
+                required: true
             mapping-file:
                 description: "Optional artifact name of the ProGuard mapping file"
                 type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-upload-to-internal-and-release.yml
```diff
@@ -15,6 +15,10 @@ on:
                 description: "Android version code for the build (used to download the universal APK from Play)"
                 type: string
                 required: true
+            build-number:
+                description: "CI build number to append to the released APK filename (e.g. 173)"
```
luhmirin-s, line 19: IIRC, we already did this and it lead to the confusing instances when a build in LT with lower number was newer because the action was executed on a different pipeline. 

Would it be feasible to add the apk build number instead or would that be too long?

meladRaouf: by APK build number you mean the version-code ?

luhmirin-s: Yeah. Or something else that can definitely tell us which APK is newer.

meladRaouf: switched to the version code. It already ends in the run number

luhmirin-s (approved):

PR #1790, opened 2026-08-17
MS-1451 Module selection rework

luhmirin-s:
[JIRA ticket](https://simprints.atlassian.net/browse/MS-1451)
Will be released in: **2027.1.0**

### Notable changes

* Re-implemented module selection screen in a reuseable module. Preserved features include: 
  * Settings lock overlay
  * Showing list of modules with clear indication of which ones are selected
  * Text search option to filter module list
  * Sync on confirm
* Changes compared to the old version:
  * Selected modules are marked with "selected checkbox" instead of chip
  * Added a toggle to filter only "selected" modules
  * "Confirm" button is enabled only when valid amount of modules is selected, otherwise and error messages is displayed

image
image
image
image

* Additionally to the functional rework this PR is an example of potential MVI architecture that we might use in the future refactoring:
  * `ModuleSelectorMvi.kt` contains definitions of the UI state, action and side effects present on the screen
  * `ModuleSelectorViewModel.kt` handles the UI actions in **vm.onAction()** and updates the state or emits side-effects accordingly
  * `ModuleSelectorFragment.kt` collects **vm.state** shared flow and renders the layout, collects **vm.effects** for side effects and sends user actions from view listeners to VM via **vm.onAction()**.

### Testing guidance

* Module selection testiny cases + updated Appium tests.

### Additional work checklist

* [ ] Effect on other features and security has been considered
* [ ] Design document marked as "In development" (if applicable)
* [ ] External (Gitbook) and internal (Confluence) Documentation is up to date (or ticket created)
* [ ] Test cases in Testiny are up to date (or ticket created)
* [ ] Other teams notified about the changes (if applicable)

PR #1831, opened 2026-09-15
[Draft] MVI Architecture Rework - Module selection screen

alexandr-simprints:
# Background

The module selector screen updated [here](https://github.com/Simprints/Android-Simprints-ID/pull/1790) serves as a good example for a complete MVI architecture proposal. It contained the reactiveness on the UI side, defined user actions and effects, and mapped the data-layer entities to UI-specific primitives.

This PR is a demonstration of a vision of how we can achieve a uniformed structure across all UI classes and its ViewModels during the Compose migration.

#  What this PR does

This PR introduces a lightweight unidirectional architecture for the presentation layers, and proposes a reusable foundation, for future screens in the project. The core idea is a strict one-way flow: 

- Data arrives from repositories into a typed state object
- That state is mapped into a UI model that the fragment observes
- All user input enters through a single channel as typed actions
- Single time events (_effects, i.e.: a toast or dialog popup_) can be emitted to the ovservers from the ViewModel

The implementation is deliberately minimal: there is no third-party libraries, no code generation, and no annotation processing.

# Fragment and reactiveness

The fragment observes a single `StateFlow<ModuleSelectorUiModel>` from the ViewModel and re-renders whenever a new model arrives. Every visual element on the screen is derived exclusively from that model. The fragment never reads from the ViewModel directly to make a decision; it only receives and renders.

# UI Model for render

`ModuleSelectorUiModel` is mapped from the state in `ModuleSelectorUiModelMapper` every time the data in the state changes. The mapper takes the current `ModuleSelectorState` and projects it into the UI primitives for the fragment.

# Data State

`ModuleSelectorState` is the single source of truth. It holds the raw data as received from the data layer (_repositories, managers, etc._).

All changes to this state go through a single queued channel defined as `Channel<(DataState) -> DataState>`. A single coroutine consumes this channel one item at a time, applies the transform, and immediately pushes the new UI model to the fragment. This ordering guarantee that no race condition appears in the state management.

# User interactions

User interactions are defined in `ModuleSelectorAction` - a sealed interface with one subclass per intent: `ModuleClicked`, `SearchQueryChanged`, etc.. The fragment dispatches these via `viewModel.onAction()` , and actions are processed sequentially through a channel in the same ViewModel. The actions can then trigger a state update or an emission of an effect.

# One-time events

Things that should happen once are emitted as `ModuleSelectorEffects` through a `SharedFlow`. The fragment collects these in the same lifecycle-aware scope as the model and handles them imperatively. Effects are explicitly separated from state so they are never accidentally re-applied when the fragment reconnects after a configuration change.


feature/module-selector/src/main/java/com/simprints/feature/moduleselector/model/ModuleSelectorUiModel.kt
```diff
@@ -0,0 +1,14 @@
+package com.simprints.feature.moduleselector.model
+
+import com.simprints.feature.moduleselector.ModuleSelectorState.SelectionError
+import com.simprints.feature.moduleselector.adapter.ModuleSelectorItem
+import com.simprints.feature.moduleselector.arch.UiModel
+
+internal data class ModuleSelectorUiModel(
```
luhmirin-s, line 7: Could you elaborate on the benefits of this in detail? More I look at it, more it feels like all it does is moves some data from VM fields to a data class and duplicates the state with very minor changes.

feature/module-selector/src/main/java/com/simprints/feature/moduleselector/model/ModuleSelectorUiModelMapper.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val tokenizationProcessor: TokenizationProcessor,
+) : StateToModelMapper<ModuleSelectorState, ModuleSelectorUiModel> {
+    private val untokenizedModulesCache = mutableMapOf<TokenizableString, String>()
+
+    override fun mapStateToModel(state: ModuleSelectorState): ModuleSelectorUiModel {
+        val filteredModules = state.allModules
+            .filter { !state.onlySelected || it.isSelected }
+            .filter { state.query.isBlank() || resolveDisplayName(it.name, state).contains(state.query, ignoreCase = true) }
+            .map { module ->
+                ModuleSelectorItem.Module(
```
luhmirin-s, line 20: This is not an efficient way to handle lists. It re-creates every item instance on any change in the state, even ones that are not affecting module items, e.g. when screen is unlocked.

feature/module-selector/src/main/java/com/simprints/feature/moduleselector/model/ModuleSelectorUiModel.kt
BurningAXE: I'm also curious to know why we need two layers between data and UI? Can't we directly convert the data to a model that maps to UI?
The downside is obvious - lots more boilerplate even for simple screens so it should be well justified.

